PR #77, opened 2015-05-21
Remove heap manipulation from fast path.

runderwo:
This reduces the CPU usage of fping somewhat (measured using top).  It helps with many fpings competing for a CPU, and/or small (1ms tested) ping intervals.


schweikert:
Did you run valgrind? Because this is not the only place where global memory isn't freed when exiting, and I still wouldn't call it a memory leak. So for the sake of uniformity, could you remove the atexit call?

runderwo:
done

schweikert:
Thanks! Could you also rebase to the develop branch, or merge the change? I merged before another change that conflicts with this one.

runderwo:
done

PR #82, opened 2015-10-09
Feature/unify

LynxChaus:
Repair checksumming


schweikert:
That's great. Thanks for your contribution!

PR #88, opened 2015-11-03
Remove trailing whitespaces

LynxChaus:


schweikert:
Thanks, but the trailing whitespaces don't bother me so much and causes noise in diffs.

Issue #87, opened 2015-10-28
Build from src


reinerotto:
autoreconf -i
missing, when cloning from git, so no ./configure available.

xtaran:
`autoreconf -i` is not missing. Have a look at [autogen.sh](https://github.com/schweikert/fping/blob/develop/autogen.sh).

`autogen.sh` is a common place to put commands necessary to generate the files necessary for building.

It's very common practice (and in general considered good practice) to not include any generated file in a git repository, only in the released tar balls.

I'd be happy if fping's git repository would stay as clean as it is with regards to generated files.

reinerotto:
It is old practice, to either to supply correct docs, or none. Docs, which assume the reader to know already part of the proedure, are not very helpful.
I refere to
"If you want to install fping from source, proceed as follows:
    Run ./configure with the correct arguments (see: ./configure --help)
    Run make; make install"

Which should better read

"If you want to install fping from source, proceed as follows:
autoreconf -i
Run ./configure with the correct arguments (see: ./configure --help)
...."

PR #89, opened 2015-11-10
Be more precise when clarifying issue reported in #87

xtaran:
"getting the source from github" could also be interpreted as
downloading released source tar balls from
https://github.com/schweikert/fping/releases -- where it doesn't apply.

The actual relevant difference is not "github or not github" but
"release tar ball or git clone".

### DIFF
--- a/README
+++ b/README
@@ -20,7 +20,7 @@ Installation
 If you want to install fping from source, proceed as follows:
 
 0. Run ./autogen.sh
-   (only if you got the source from github)
+   (only needed if you got the source code by cloning a git repository)
 
 1. Run ./configure with the correct arguments
    (see: ./configure --help)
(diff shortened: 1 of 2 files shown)


schweikert:
Thanks for the patch, but actually, it does apply to the source tar balls from https://github.com/schweikert/fping/releases, so maybe the text is good as is?

xtaran:
Urgs. Indeed.

So why is there both https://github.com/schweikert/fping/releases and https://github.com/schweikert/fping/tags if https://github.com/schweikert/fping/releases doesn't contain the same tar balls as http://www.fping.org/dist/? I actually expected that the tar balls on https://github.com/schweikert/fping/releases are the same as on http://www.fping.org/dist/. Anything else doesn't make sense IMHO.

PR #95, opened 2016-04-08
Issue compliling on RHEL/Centos 7

jbackman:
warning: compiling 'fping.c' with per-target flags requires 'AM_PROG_CC_C_O' in 'configure.ac'

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -37,6 +37,7 @@ AC_CONFIG_HEADERS([config.h])
 dnl Checks for programs.
 
 AC_PROG_CC
+AM_PROG_CC_C_O
 AC_PROG_CPP
 AC_PROG_INSTALL
 


bsilva-bentlabs:
Any idea how to fix this?

I have some software that requires fping, but it still doesn't compile under RHEL or CentOS 7.

Result on CentOS 7:
$  ./autogen.sh
configure.ac:31: installing './config.guess'
configure.ac:31: installing './config.sub'
configure.ac:32: installing './install-sh'
configure.ac:32: installing './missing'
automake: warnings are treated as errors
src/Makefile.am:16: warning: compiling 'fping.c' with per-target flags requires 'AM_PROG_CC_C_O' in 'configure.ac'
src/Makefile.am: installing './depcomp'
autoreconf: automake failed with exit status: 1

Thanks,
Brad

bsilva-bentlabs:
Doh!

Just saw that this was a commit message.  I got here via Google and thought it was a bug report.

I added the line in this commit to my configure.ac and that fixed my issue.

Cheers,
Brad

PR #96, opened 2016-04-14
Added support for coloured output.

SA0TAY:
Was handy in a situation we had last night. Might be again for someone else.

### DIFF
--- a/doc/fping.pod
+++ b/doc/fping.pod
@@ -118,6 +118,10 @@ Print usage message.
 The minimum amount of time (in milliseconds) between sending a ping packet
 to any target (default is 25). 
 
+=item B<-k>
+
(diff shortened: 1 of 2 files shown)


xtaran:
Hi David,

I didn't see any related commit, but you closed this without a comment. So is this feature not wanted or was the closing a mistake?

schweikert:
Yes, sorry, a mistake. Thanks for noticing!

Eideen:
It does looks like there is a issue when using  `-k` and `-q`.
Where it will not produce any output.
Example: `./fping -g 192.168.2.1/26 -k -q`

auerswal:
I'll convert this PR to a "draft".

PR #102, opened 2016-09-02
Ignore network and broadcast for cidrs /31 and /32

martintopholm:
fping has previously been strict about network address and broadcast
addreses. In the commit this is loosened a bit to permit fping to ping
cidrs with prefix length 31, which are commonly used for link ranges
(for newer equipment) and length 32 which denotes a single ip address.

For prefix lengths 31 and 32 fping will simply ignore network address
and broadcast addresses, and consider all addresses in range as a
target.


schweikert:
Thanks!

PR #106, opened 2016-11-01
minor netdata improvements

ktsaou:
1. converted netdata charts to `area` charts

    I think charts like this:

     image

    are nicer, than this:

    image

2. added netdata to `trace_flag` output

3. fixed compiler warning about `setuid()` return value being ignored, by calling `perror()` on `setuid()` errors.

4. fix 100% cpu consumption introduced with https://github.com/schweikert/fping/commit/7c7e007502a3a2bd435a9e6b797387dbea8b992c

### DIFF
--- a/src/fping.c
+++ b/src/fping.c
@@ -369,7 +369,8 @@ int main( int argc, char **argv )
 
     if((uid = getuid())) {
         /* drop privileges */
-        setuid( getuid() );
+        if(setuid( getuid() ) == -1)
+            perror("cannot setuid");
     }
 
     prog = argv[0];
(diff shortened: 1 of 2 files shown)


src/fping.c
```diff
@@ -1099,6 +1101,8 @@ void main_loop()
         /* Make sure we don't wait too long, in case a report is expected */
         if( report_interval && ( loop_flag || count_flag ) ) {
                wait_time_next_report = timeval_diff ( &current_time, &next_report_time );
+               if(wait_time_next_report < 0)
+                    wait_time_next_report = -wait_time_next_report;
```
schweikert: This doesn't look right. Why did you add this?

schweikert: I have added a commit that should fix this.

ktsaou: ok, removed it.

ktsaou:
perfect! thanks!

PR #107, opened 2016-11-19
Portability Fixes discovered on Solaris 10, but should be generally applicable

illumino:
David,

These two commits are both trivial and hopefully appropriately explained in the commit messages.

Tested on Solaris 10 Update 8, Solaris 10 Update 11 and Solaris 11.2 with locally built gcc 4.8.5 and gcc 4.9.3 for both 32-bit and 64-bit compiles. I only have access to X86 Solaris Systems (not SPARC).

Regards,
Peter

### DIFF
--- a/src/fping.c
+++ b/src/fping.c
@@ -365,6 +365,8 @@ int main( int argc, char **argv )
     int tos = 0; 
     HOST_ENTRY *cursor;
 
+    prog = argv[0];
+
     s = open_ping_socket(ping_data_size);
 
     if((uid = getuid())) {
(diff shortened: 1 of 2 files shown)


illumino:
David,

The CI checks have failed as they are expecting the program name to be "(null)", and one of the two patches (commits) changes that, so the CI tests will have to be updated. Sorry I did my development on fping-3.13 tar ball and just patched and committed the changes in github. It looks like two changes will be required to "ci/test-11-nopriv.pl", sorry but I don't have the tools to develop and test such changes.

Peter

schweikert:
Thanks!

PR #124, opened 2018-02-16
Return -1 when socket_can_read times out.

### LINKED ISSUE #125: socket_can_read doesn't work when the socket's fd is 0
First of all, thanks for fping. It's a great piece of software!

Yesterday I saw it fail when called from a script. The problem was that [socket_can_read](https://github.com/schweikert/fping/blob/16aaa609cd5873353a0492eb9f2566ab82e84b93/src/fping.c#L1677) doesn't work when the socket's file descriptor is 0, because it uses this value to signal a [timeout](https://github.com/schweikert/fping/blob/16aaa609cd5873353a0492eb9f2566ab82e84b93/src/fping.c#L1719).

This behaviour can be reproduced fairly easily:
```                                                                                                                                                                                                                                          
# bash -c 'exec 0<&-; src/fping 127.0.0.1'
127.0.0.1 is unreachable
```
                                                                                                                                                                                                                                      
I've submitted a pull request (#124) that fixes the problem:
```                                                                                                                                                                                                                                         
# bash -c 'exec 0<&-; src/fping 127.0.0.1'
127.0.0.1 is alive
```

Best regards.

nramon:
0 is a valid file descriptor. Use -1 instead to signal a timeout. Fixes #125.

### DIFF
--- a/src/fping.c
+++ b/src/fping.c
@@ -1716,7 +1716,7 @@ select_again:
 #endif
     }
 
-    return 0;
+    return -1;
 }
 
 int receive_packet(int socket,
@@ -2005,7 +2005,7 @@ int wait_for_reply(long wait_time)
         to.tv_usec = 0;
     }
     s = socket_can_read(&to);
-    if (s == 0) {
+    if (s == -1) {
         return 0; /* timeout */
     }
 


schweikert:
Thanks! It will be included in fping 4.1

PR #132, opened 2018-08-31
Fix FreeBSD ipv6 build

gsnw:
Source: https://svnweb.freebsd.org/ports/head/net/fping/files/patch-configure.ac?revision=473606&view=markup
Fix build fping with ipv6 support on FreeBSD


schweikert:
Thanks!

PR #134, opened 2018-09-18
Fix build with --disable-ipv6

Polynomial-C:
Otheriwse build breaks with:

fping.c:690:24: error: ‘socket6’ undeclared (first use in this function);
did you mean ‘socket4’?


schweikert:
Thanks!

PR #136, opened 2018-09-30
Move to 1us resolution for latency measurements

### LINKED ISSUE #101: Change resolution to 1us rather than 10us
Currently the minimum resolution for fping is 10us.  Ideally that would be lowered to 1us.  This is especially useful for modern low latency devices.  Ultimately, I'm sure measurements under 1us will be desired.

tycho:
A few fixes:

- The `-z` debug flag was implemented but not accepted by getopt_long. I added it.
- The `-z` debug flag only accepted a decimal argument, and since the argument is a bitfield, I made it possible to take a hex argument, i.e. `0xffff` enables all debug bits.

And the most important part:

- Use 1us resolution for latency measurements. This fixes #101.

### DIFF
--- a/ci/test-05-options-c-e.pl
+++ b/ci/test-05-options-c-e.pl
@@ -41,8 +41,8 @@ $cmd->stderr_like(qr{localhost : \d\.\d+ \d\.\d+
 {
 my $cmd = Test::Command->new(cmd => "fping -D -c 2 -p 100 127.0.0.1");
 $cmd->exit_is_num(0);
-$cmd->stdout_like(qr{\[\d{10}\.\d+\] 127\.0\.0\.1 : \[0\], 84 bytes, \d\.\d+ ms \(\d\.\d+ avg, 0% loss\)
-\[\d{10}\.\d+\] 127\.0\.0\.1 : \[1\], 84 bytes, \d\.\d+ ms \(\d\.\d+ avg, 0% loss\)
+$cmd->stdout_like(qr{\[\d+\.\d+\] 127\.0\.0\.1 : \[0\], 84 bytes, \d\.\d+ ms \(\d\.\d+ avg, 0% loss\)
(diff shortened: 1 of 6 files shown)


tycho:
Looks like `SO_TIMESTAMPNS` always uses `CLOCK_REALTIME` instead of `CLOCK_MONOTONIC`, so the numbers don't line up:

```
timestampns: 1538342854 527430214
clock_gettime: 5419 195062250
timestampns: 1538342855 528457016
clock_gettime: 5420 196056278
```

Bummer. `CLOCK_MONOTONIC` tends to have better quality. I'll make it use `CLOCK_MONOTONIC` only if the builder has opted out of using the `SO_TIMESTAMPNS` feature.

luckman212:
@tycho Any time to check into the build errors so your PR can be merged?

luckman212:
@tycho Any time to check into the build errors so your PR can be merged?

luckman212:
@tycho Any time to check into the build errors so your PR can be merged?

luckman212:
@tycho Any time to check into the build errors so your PR can be merged?

tycho:
Well, I assume you're referring to the Travis CI errors.

I've fixed those now:
- I was enqueuing the targets incorrectly. I used `>` instead of `>=` in ev_enqueue, which caused targets to be pushed to the front instead of back.
- The netdata test was broken because it used a regular expression for the result expecting a specific set of digits, but we're printing a higher resolution result now. I need to verify that it renders correctly in netdata now.

tycho:
The netdata output mode now appears to work properly. Helps if I tell it to use the right divisor.

schweikert:
Hi Steven,

If you agree, I think we could move ahead with merging the code and target a fping 5.0 some times in the future. I will also create a branch for further 4.x releases. What do you think? Is the patch ready for merging?

tycho:
> Hi Steven,
> 
> If you agree, I think we could move ahead with merging the code and target a fping 5.0 some times in the future. I will also create a branch for further 4.x releases. What do you think? Is the patch ready for merging?

Yeah, I think it's ready to go.

Issue #99, opened 2016-08-05
request new option to support connectivity tests


rlhamil:
I'd like to see a new option to change the return code behavior from the default, so that if ANY host pings succeeded, the return code would be zero, even if not all succeeded or some IP lookups failed.  Alternatively, a more generic option could take a numeric argument, and in addition to the above behavior, would randomly pick that number (after address resolution?) of hosts to ping out of a larger (potentially much larger) list.  That would allow scripts doing such things to be used by many people, yet spread the load to the point that they weren't effectively causing a DoS attack on any one host.

Say the new option was -x (not suggesting that, although it is unused); for the simpler implementation, something like
fping -x -q host1 host2 host3
would return zero if any of the three was resolvable and responded.  For the more complicated approach,
fping -x 2 -q host1 ... host99
would pick two hosts out of those of the 99 that resolved or didn't need name resolution, and if either responded, would return zero.

filbo:
I came here to request this, so, since the request already exists I'll just "me too".

Application: in my case, I want to bring up NIS to my corporate network when and only when I have IP connectivity to at least one of the N YP servers.  Connectivity could be via direct LAN, on-campus wifi, or VPN from any other sort of network anywhere.  And of course Everything Sucks when YP is "up" but connectivity is severed, so it has to die quickly when connectivity goes down; and rise from the dead reasonably quickly when connection is established.

fping, or even regular ping, already provides the needed functionality.  A "tell me if *any* of these hosts are reachable" exit status is just a triviality to make scripting slightly easier.  But still.

Seems like a generic "cause service P to be alive whenever any host in set Q is reachable (and not-alive when not, with some sort of tunable latency / hysteresis)" daemon should already exist; maybe I just need someone to remind me of its name?

rlhamil:
Yes, --reachable does what I had in mind.

The DIAGNOSTICS portion of the man page currently (at least for what I have installed, which may not be the very latest) says
       Exit status is 0 if all the hosts are reachable, 1 if some hosts were
       unreachable, 2 if any IP addresses were not found, 3 for invalid command
       line arguments, and 4 for a system call failure.

Perhaps it should say
       Exit statius is 0 if all the hosts are reachable or the number of hosts specified by --reachable are reachable, [etc]

just so one doesn't have to read the entire man page to discover the exception.

PR #138, opened 2018-10-17
Adding -x option which allows user to specify a threshold and check if atleast those many hosts are active

deepak0004:
The feature as asked by #99

### DIFF
--- a/src/fping.c
+++ b/src/fping.c
@@ -1195,7 +1195,7 @@ void main_loop()
                     h->ev_type = EV_TYPE_PING;
                     h->ev_time.tv_sec = last_send_time.tv_sec;
                     h->ev_time.tv_usec = last_send_time.tv_usec;
-         	        timeval_add(&h->ev_time, perhost_interval);
+         	      timeval_add(&h->ev_time, perhost_interval);
                     ev_enqueue(h);
                 }
                 /* Count mode: schedule timeout after last ping */
@@ -1281,13 +1281,6 @@ void main_loop()
                 timeval_add(&next_report_time, report_interval);
         }
     }
-    if (reachable_flag) {
-        if ((num_hosts-num_unreachable >= reachable)) {
-            printf(">=%d hosts are reachable\n", reachable);
-        } else {
-            printf("<%d hosts are reachable\n", reachable);
-        }
-    }
 }
 
 /************************************************************
@@ -1326,6 +1319,14 @@ void finish()
 
                 printf("\n");
             }
+        }
+    }
+
+    if (reachable_flag) {
+        if ((num_hosts-num_unreachable >= reachable)) {
+            printf(">=%d hosts are reachable\n", reachable);
+        } else {
+            printf("<%d hosts are reachable\n", reachable);
         }
     }
 
@@ -1904,7 +1905,7 @@ int decode_icmp_ipv4(
     }
 
     *id = ntohs(icp->icmp_id);
-    *seq = ntohs(icp->icmp_seq);    
+    *seq = ntohs(icp->icmp_seq);
 
     return 1;
 }
--- a/src/fping.c
+++ b/src/fping.c
@@ -239,7 +239,7 @@ int socket6 = -1;
 int hints_ai_family = AF_UNSPEC;
 #endif
 
-unsigned int debugging = 0;
+unsigned int debugging = 1;
 
 /* times get *100 because all times are calculated in 10 usec units, not ms */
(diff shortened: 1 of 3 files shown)


deepak0004:
Hi @schweikert, 
Could you review this PR?

src/fping.c
```diff
@@ -1323,6 +1332,15 @@ void finish()
     if (stats_flag)
         print_global_stats();
 
+    if (reachable_flag) {
+        if ((num_hosts-num_unreachable >= reachable)) {
+            printf(">=%d hosts are reachable\n", reachable);
+        } else {
+            printf("<%d hosts are reachable\n", reachable);
+        }
+        exit(0);
```
schweikert: I think that for this to be useful it should return a different exit code, depending on failure or success (and hopefully scripts using this won't rely on the output...)

src/fping.c
```diff
@@ -1323,6 +1332,15 @@ void finish()
     if (stats_flag)
         print_global_stats();
 
+    if (reachable_flag) {
+        if ((num_hosts-num_unreachable >= reachable)) {
+            printf(">=%d hosts are reachable\n", reachable);
```
schweikert: Should you maybe also include the number of reachable hosts in the output?

src/fping.c
deepak0004: I added different exit codes, 0 for success (if >= N hosts are reachable), and 1 for failure (if < N hosts are reachable).

src/fping.c
deepak0004: I have printed the number of reachable hosts instead of the given statement.

src/fping.c
```diff
@@ -1323,6 +1332,16 @@ void finish()
     if (stats_flag)
         print_global_stats();
 
+    if (reachable_flag) {
+        if ((num_hosts-num_unreachable >= reachable)) {
+            printf("Number of reachable hosts: %d\n", num_hosts-num_unreachable);
```
schweikert: I was thinking of maybe something like:

```suggestion
            printf("Enough hosts reachable (required: %d, reachable: %d)\n", reachable, num_hosts-num_unreachable);
```

src/fping.c
```diff
@@ -1323,6 +1332,16 @@ void finish()
...
+            exit(0);
+        } else {
+            printf("<%d hosts are reachable\n", reachable);
```
schweikert: Here too:

```suggestion
            printf("Not enough hosts reachable (required: %d, reachable: %d)\n", reachable, num_hosts-num_unreachable);
```

src/fping.c
```diff
@@ -285,7 +285,7 @@ struct timezone tz;
 /* switches */
 int generate_flag = 0; /* flag for IP list generation */
 int verbose_flag, quiet_flag, stats_flag, unreachable_flag, alive_flag;
-int elapsed_flag, version_flag, count_flag, loop_flag, netdata_flag;
+int elapsed_flag, version_flag, count_flag, loop_flag, netdata_flag, reachable_flag;
```
schweikert: Only one variable should be enough I think (i.e. the variable that contains the value of the reachable flag)

src/fping.c
```diff
@@ -248,7 +248,7 @@ unsigned int interval = DEFAULT_INTERVAL * 100;
 unsigned int perhost_interval = DEFAULT_PERHOST_INTERVAL * 100;
 float backoff = DEFAULT_BACKOFF_FACTOR;
 unsigned int ping_data_size = DEFAULT_PING_DATA_SIZE;
-unsigned int count = 1;
+unsigned int count = 1, reachable = 1;
```
schweikert: I suggest to rename "reachable" to something more descriptive like "min_reachable". Also, the default should be 0?

```suggestion
unsigned int count = 1, min_reachable = 0;
```

src/fping.c
deepak0004: Made the required modification in both test cases and the code.

src/fping.c
deepak0004: Made the required modification in both test cases and the code.

src/fping.c
deepak0004: Removed the reachable_flag variable (taking min_reachable>0 value as the flag as proposed).

src/fping.c
deepak0004: Changed the initialisation.

schweikert:
Thanks! I will revert the option to be called --reachable again instead of --min_reachable, because I think it looks nicer.

PR #141, opened 2018-11-26
Add JSON output with new --json/-J option

jasoncdavis:
Added support for a -J output flag to report ping results in JSON records - this is more desirable for programmatic consumption of the fping data

### DIFF
--- a/src/fping.c
+++ b/src/fping.c
@@ -291,6 +291,7 @@ int multif_flag, timeout_flag;
 int outage_flag = 0;
 int timestamp_flag = 0;
 int random_data_flag = 0;
+int output_json_flag;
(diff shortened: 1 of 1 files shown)


jasoncdavis:
Added support for a -J output flag to report ping results in JSON records - this is more desirable for programmatic consumption of the fping data

schweikert:
Thanks for your contribution! There are some changes that I would do before integrating it:
- Call the option --json, instead of --JSON for consistency with the other options
- Use standard output for the JSON output when in JSON mode, and stderr for everything else (as it should be also for the non-json mode, but we can't fix it there anymore)
- Add some tests for this

schweikert:
Continued in #160

PR #148, opened 2019-05-24
Corrected long option name of -s to --stats

wopfel:
The option --src is an alias for the uppercase -S.

Stumbled upon this in the man page.

When running "fping -h", the option is printed correctly:
$ -s, --stats        print final stats


schweikert:
Thanks!

Issue #147, opened 2019-05-07
fping6 -6 fails on 4.2


stromnet:
My FreeBSD smokeping setup was configured with binary name `fping6`, and smokeping 2.7.3 which always uses `-6`.
When fping 4.2 got available in FreeBSD ports (prev 4.0), fping6-probes from smokeping stopped working.

After some digging, it turns out `fping6` and `-6` is not compatible, and fails with `cant specify both -4 and -6`. 

Strictly speaking, it's would make sense to support that combination, but perhaps it is more desirable to keep this behaviour and enforce user to stop using the legacy `fping6` binary (but if so, perhaps give a better error)?

Quickfix, but not creating MR due to question above:
```
--- a/src/fping.c
+++ b/src/fping.c
@@ -435,7 +435,7 @@ int main(int argc, char** argv)
             break;
         case '6':
 #ifdef IPV6
-            if (hints_ai_family != AF_UNSPEC) {
+            if (hints_ai_family != AF_UNSPEC && hints_ai_family != AF_INET6) {
                 fprintf(stderr, "%s: can't specify both -4 and -6\n", prog);
                 exit(1);
             }
```

And the actual fix I used in the end, was to use `fping` binary instead... as described (and automatically detected) in https://github.com/oetiker/SmokePing/issues/146... So perhaps this is just a useless ticket, but this might hit someone else too :)

schweikert:
For sure it would be better to encourage users to abandon "fping6", but your proposed fix looks good to me and I think does no harm. Will you do a pull request?

stromnet:
Agree, this at least avoids sudden failures after updating. Pushed MR.

PR #149, opened 2019-06-25
Do not fail if using fping6 with -6 flag

stromnet:
Proper fix is for user to abandon fping6 but at least this makes it work.

See #147 for discussion and more details.


schweikert:
Thanks!

PR #151, opened 2019-07-07
Only increase num_alive if response is not a duplicate

brownowski:
Moved code block for num_alive increment. Should only increase num_alive if response is not a duplicate.  

In cases where multiple pings are sent, num_alive can increase past one for a host if a duplicate ping response is received while h->num_recv ==1.

### DIFF
--- a/src/fping.c
+++ b/src/fping.c
@@ -2149,6 +2149,27 @@ int wait_for_reply(long wait_time)
         sum_replies += this_reply;
         h->total_time += this_reply;
         total_replies++;
+        
+        if (h->num_recv == 1) {
+            num_alive++;
+            if (verbose_flag || alive_flag) {
+                printf("%s", h->host);
+
+                if (verbose_flag)
+                    printf(" is alive");
+
+                if (elapsed_flag)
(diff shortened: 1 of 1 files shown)


schweikert:
Can you explain why moving the block makes a difference?

brownowski:
It has been a while since I submitted this patch, but this is the problem to the best of my current knowledge.

There is a problem with VMWare virtual machines where if a NIC is part of a team, duplicate packets can show up on the interface.

This means that when fping is used from within this VM, it can end up seeing multiple responses for each ping it sends out.

While the existing code does flag that there are duplicate responses, there is an error with the code with how fping counts the number of received packets when there are duplicate responses, meaning that it can increase the counter for num_alive past the number of packets it has sent, and then cause it to flag an error in its return code because the received packets number doesn't match the sent packets.

In moving this code block, I'm attempting to prevent it from counting a received ping if it detects the ping response as a duplicate so that the number of received pings will not cause it to return an error code on exit.

A workaround of sorts is to limit the ping count to 1, which causes the program to exit before the duplicates are received and avoid the error in the return code.

PR #154, opened 2019-07-28
Fix -4 option when IPv6 disabled

pbhenson:
When IPv6 is disabled, hints_ai_family is set to AF_INET at initialization, resulting in the error "can't specify both -4 and -6" if the -4 option is supplied. If there's no IPv6, make the -4 option a no-op.


schweikert:
Thanks!

Issue #156, opened 2019-08-21
Packet loss displayed during execution is "incorrect" when timeout (-t) larger than period (-p)


bkuker:
When the timeout is larger than the period the calculated packet loss is too high.
That calculation includes all pings sent, including those that have not timed out yet.

On the following execution, running 10 trials with 20ms period and 100ms timeout:

```
bkuker@bill-system:~/fping$ fping --size=200 --period=20 --count=10 --timeout=100 www.csh.rit.edu
fping: warning: timeout (-t) value larger than period (-p) produces unexpected results
www.csh.rit.edu : [0], 228 bytes, 77.5 ms (77.5 avg, 75% loss)
www.csh.rit.edu : [1], 228 bytes, 77.6 ms (77.5 avg, 60% loss)
www.csh.rit.edu : [2], 228 bytes, 77.5 ms (77.5 avg, 50% loss)
www.csh.rit.edu : [3], 228 bytes, 77.4 ms (77.5 avg, 42% loss)
www.csh.rit.edu : [4], 228 bytes, 77.4 ms (77.5 avg, 37% loss)
www.csh.rit.edu : [5], 228 bytes, 77.5 ms (77.5 avg, 33% loss)
www.csh.rit.edu : [6], 228 bytes, 77.4 ms (77.5 avg, 30% loss)
www.csh.rit.edu : [7], 228 bytes, 77.4 ms (77.5 avg, 20% loss)
www.csh.rit.edu : [8], 228 bytes, 77.6 ms (77.5 avg, 10% loss)
www.csh.rit.edu : [9], 228 bytes, 77.4 ms (77.5 avg, 0% loss)

www.csh.rit.edu : xmt/rcv/%loss = 10/10/0%, min/avg/max = 77.8/77.9/78.0
```

Note that no packets are lost, but by the time response [0] is returned, 77ms after request[0] was sent, requests [1], [2], and [3] have already been sent, and the responses are in flight but have not been received yet.
These requests are counted against packet loss, resulting in a 75% loss being reported.

I think the correct behavior should be that outstanding responses *within the timeout period* should not be counted as either lost or received.

I'm working on a simple fix I will share if you'd like to fix this behavior.

bkuker:
BTW This is my best guess about what #142 means about "unexpected results" because everything else seems as expected.

schweikert:
fixed in #193

PR #157, opened 2019-08-21
Do not count packets that are still within their timeout window in packet loss calculation.

bkuker:
Fix for issue #156 .

Adds a circular buffer of ping sent times `int *window` of length at least ceil(timeout/period). Entries are set to the send time when the ping is sent, and cleared (set to int_min) when a response is received.

During the per-receive stats print out `www.csh.rit.edu : [3], 228 bytes, 77.4 ms (77.5 avg, 42% loss)` these sent times are compared to the current time + the timeout. Any requests that are still unanswered, but within their timeout window, are excluded from the packet loss calculation.

Please excuse all the extra `fprintf`s, I will clean out my debugging output as I finalize this PR, if you indicate that it might be accepted.

Thank you!

### DIFF
--- a/src/fping.c
+++ b/src/fping.c
@@ -2250,8 +2250,8 @@ int wait_for_reply(long wait_time)
             }
 #endif
         }
-        if ( expected < 0 )
-                expected = 0;
+        if ( expected < 1 )
+                expected = 1;
 
         if (h->num_recv <= expected) {
             printf("%d%% loss)",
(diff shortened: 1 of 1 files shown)


schweikert:
Thanks for the patch! I don't have much time to look at this right now, but I certainly will. I worry a bit about performance aspects, so I will need to verify the cost of this.

schweikert:
This is now fixed with a different implementation in #193

PR #163, opened 2019-12-05
Fail on interface binding errors

### LINKED ISSUE #162: fping does not fail if  interface binding (-I) fails
This gives wrong information as the destination is not reachable over the intended link:
```
# ip addr show eth8
Device "eth8" does not exist.
#
# fping -I eth8 8.8.8.8
binding to specific interface (SO_BINTODEVICE): No such device
binding to specific interface (SO_BINTODEVICE), IPV6: No such device
8.8.8.8 is alive
# 
```

kbucheli:
This makes fping fail if the interface binding does not work, fixes #162


schweikert:
Thanks Konrad!

PR #169, opened 2020-02-03
Add Windows support by add a head file from FREEBSD

willowyoung:
When I compiling these code in Cygwin on Windows.
There'are some errors caused by icmp struct undeclared and some constants undefined.
And I found that add a standard <netinet/ip_icmp.h> can solved the problem.
So I copy one from FREEBSD 
https://github.com/freebsd/freebsd/blob/master/sys/netinet/ip_icmp.h .
And it works well.


willowyoung:
...
coverage/coveralls — Coverage decreased (-0.2%) to 79.225%
...
OK，let me check it out.

PR #173, opened 2020-02-27
Linux unprivileged ping support

tycho:
Right now if you try to use unprivileged ping (i.e. via `SOCK_DGRAM`) on the "develop" branch, it doesn't see the response packets.

There are a few problems this PR fixes:
- The "ident" value cannot be specified by unprivileged processes, so we have to get it from the kernel via `getsockname`. This change is necessary in order for it to see the ident fields line up and accept the responses.
- When using `SOCK_DGRAM` for IPv4 unprivileged ping on Linux, there is no `struct ip` at the beginning of the message read by `recvmsg`. So we have to handle that situation specially. First, we need to account for the size of the missing header when logging the responses (since it *is* data sent/received over the wire, but it's not visible to the userspace process). Second, we need to avoid treating the start of the buffer as `struct ip`.
- The `test-11-nopriv` test will fail on Linux if the user is allowed to do unprivileged pings. We can check if the effective GID is within the range specified in the sysctl `net.ipv4.ping_group_range` and know in advance whether the ping is expected to work or not. If it is expected to work, then we should skip the `test-11-nopriv` test.

### DIFF
--- a/ci/test-11-nopriv.pl
+++ b/ci/test-11-nopriv.pl
@@ -1,5 +1,6 @@
 #!/usr/bin/perl -w
 
+use English;
 use Test::Command;
 use Test::More;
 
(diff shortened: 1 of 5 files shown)


schweikert:
Looks good, thank you!

PR #179, opened 2020-04-18
-L, --line-buffered Use line buffering on output

bg6cq:
Buffering mode can be change to line-buffered when the output of fping was piped to other program.


schweikert:
I wonder if maybe we could always use line buffering. Does it matter performance-wise?

I suggest to make stdout line buffered, without option, and later add an option for it if we hear complaints about performance regression.

bg6cq:
Great, I agree to always use line buffering.

PR #185, opened 2020-06-12
Add SIGQUIT summary support similar to ping

laddp:
ping supports printing out an intermediate summary of results and resuming via `SIGQUIT` (`ctrl-\`). fping currently prints out summary statistics only for hosts responding.  This allows a summary to be output of all hosts without terminating the run.

Sample output:
```
fping -g 192.168.42.210 192.168.42.212 -l
192.168.42.211 : [0], 84 bytes, 1.00 ms (1.00 avg, 0% loss)
192.168.42.211 : [1], 84 bytes, 2.40 ms (1.70 avg, 0% loss)
192.168.42.211 : [2], 84 bytes, 0.93 ms (1.44 avg, 0% loss)
192.168.42.211 : [3], 84 bytes, 1.31 ms (1.41 avg, 0% loss)
ICMP Host Unreachable from 192.168.42.150 for ICMP Echo sent to 192.168.42.212
ICMP Host Unreachable from 192.168.42.150 for ICMP Echo sent to 192.168.42.212
ICMP Host Unreachable from 192.168.42.150 for ICMP Echo sent to 192.168.42.212
ICMP Host Unreachable from 192.168.42.150 for ICMP Echo sent to 192.168.42.212
ICMP Host Unreachable from 192.168.42.150 for ICMP Echo sent to 192.168.42.210
ICMP Host Unreachable from 192.168.42.150 for ICMP Echo sent to 192.168.42.210
ICMP Host Unreachable from 192.168.42.150 for ICMP Echo sent to 192.168.42.210
ICMP Host Unreachable from 192.168.42.150 for ICMP Echo sent to 192.168.42.210
192.168.42.211 : [4], 84 bytes, 1.03 ms (1.33 avg, 0% loss)
^\
[18:56:16]
192.168.42.210 : xmt/rcv/%loss = 5/0/100%
192.168.42.211 : xmt/rcv/%loss = 5/5/0%, min/avg/max = 0.93/1.33/2.40
192.168.42.212 : xmt/rcv/%loss = 4/0/100%
192.168.42.211 : [5], 84 bytes, 1.12 ms (1.29 avg, 0% loss)
^C
192.168.42.210 : xmt/rcv/%loss = 6/0/100%
192.168.42.211 : xmt/rcv/%loss = 6/6/0%, min/avg/max = 0.93/1.29/2.40
192.168.42.212 : xmt/rcv/%loss = 6/0/100%
```

### DIFF
--- a/src/fping.c
+++ b/src/fping.c
@@ -252,6 +252,8 @@ int socket6 = -1;
 int hints_ai_family = AF_UNSPEC;
 #endif
 
+volatile sig_atomic_t status_snapshot = 0;
+
 unsigned int debugging = 0;
 
 /* times get *100 because all times are calculated in 10 usec units, not ms */
@@ -329,6 +331,7 @@ void print_per_system_splits(void);
 void print_netdata(void);
 void print_global_stats(void);
 void main_loop();
+void sigstatus();
 void finish();
 char* sprint_tm(int t);
 void ev_enqueue(HOST_ENTRY* h);
@@ -1027,6 +1030,7 @@ int main(int argc, char** argv)
 #endif
 
     signal(SIGINT, finish);
+    signal(SIGQUIT, sigstatus);
 
     gettimeofday(&start_time, NULL);
     current_time = start_time;
@@ -1302,6 +1306,11 @@ void main_loop()
 
         gettimeofday(&current_time, NULL);
 
+        if (status_snapshot) {
+            status_snapshot = 0;
+            print_per_system_splits();
+        }
+
         /* Print report */
         if (report_interval && (loop_flag || count_flag) && (timeval_diff(&current_time, &next_report_time) >= 0)) {
             if (netdata_flag)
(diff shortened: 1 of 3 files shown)


src/fping.c
```diff
@@ -252,6 +252,8 @@ int socket6 = -1;
 int hints_ai_family = AF_UNSPEC;
 #endif
 
+volatile int status_snapshot = 0;
```
schweikert, line 255: I think "volatile" is wrong here. It is a qualifier reserved for embedded programs where the variable's value can change without the compiler knowing (memory-mapped IO, etc.).

laddp: Values changed in signal handlers also fall in that class - the variable may be changed outside of the normal flow of execution that the compiler expects.  Signal handlers are processed asynchronously from the main flow of execution.

schweikert: You are right, I learned something :) But then, shouldn't it be of type sig_atomic_t ?

laddp: Makes sense to be cross-platform correct (I think sig_atomic_t == int o most platforms, but TIL something too :) )  I'll update the PR and add it to this commit.

laddp: Ok updated to use `sig_atomic_t`.

Think I got the whole squash commit thing right...

schweikert:
Thanks!

PR #186, opened 2020-07-01
feature: single request loss rate

qiangmzsx:
Recently, I found out that when I use `fping -l` to do network monitoring, loss is the most common problem in this case. The proportion of loops of loss, which is `(h->num_sent - h->num_recv) * 100`. If fping runs for a long time in a monitoring scenario, up to a million times, and there is a network Packet loss when unavailable is difficult to detect through monitoring, as it accounts for a very small percentage.

I add the option `fping -L` to provide the percentage of loss for a single request, i.e. loss only counts the percentage of loss for a single request.
demo:
```
~/Desktop/sbin » ./fping -L -l 192.168.0.54                                                                                                                        ouerqiang@bogon
192.168.0.54 : [0], 84 bytes, 5.20 ms (5.20 avg, 0% loss)
... ...
192.168.0.54 : [23], 84 bytes, 214 ms (30.4 avg, 0% loss)
192.168.0.54 : [24], 84 bytes, 357 ms (43.5 avg, 0% loss)
192.168.0.54 : [37], 84 bytes, 106 ms (45.9 avg, 100% loss)
... ... 
192.168.0.54 : [44], 84 bytes, 5.39 ms (30.7 avg, 0% loss)
192.168.0.54 : [45], 84 bytes, 5.36 ms (30.1 avg, 0% loss)
192.168.0.54 : [51], 84 bytes, 6.67 ms (29.6 avg, 100% loss)
192.168.0.54 : [52], 84 bytes, 4.32 ms (26.4 avg, 0% loss)
```

### DIFF
--- a/src/fping.c
+++ b/src/fping.c
@@ -299,7 +299,7 @@ struct timeval next_report_time; /* time next -Q report is expected */
 /* switches */
 int generate_flag = 0; /* flag for IP list generation */
 int verbose_flag, quiet_flag, stats_flag, unreachable_flag, alive_flag;
-int elapsed_flag, version_flag, count_flag, loop_flag, netdata_flag;
+int elapsed_flag, version_flag, count_flag, loop_flag, netdata_flag, loss_flag;
 int per_recv_flag, report_all_rtts_flag, name_flag, addr_flag, backoff_flag, rdns_flag;
 int multif_flag, timeout_flag;
 int outage_flag = 0;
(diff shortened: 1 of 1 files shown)


schweikert:
I don't see the usefulness of such an option and the output actually looks confusing to me. I will close, but feel free to reopen with more information about why this is needed.

qiangmzsx:
Network monitoring using fping will be long running. According to the current loss algorithm `(h->num_sent - h->num_recv) * 100`, will result in a very low variation of the collected bosses. Or the decay is very slow, imagine that `-l` has cycled 1,000,000 times, and there is a packet loss, then the loss. Only `(1000000-999999)/100000*100%` = 0.0001%. This is a difficult threshold to set for monitoring.
So adding the single-count request packet loss option can avoid similar situations.

Issue #184, opened 2020-06-07
output not send to stdout, with  `-C` `-c`


Eideen:
If  `-C`  or `-c` is used the output is not what i expected.

I expecte the result to be sent to stdout.

### CMD: `TEST=$(fping 192.168.2.1 -C3) ;echo; echo $TEST`
Result
````bash

192.168.2.1 : 0.54 0.37 0.41

192.168.2.1 : [0], 84 bytes, 0.54 ms (0.54 avg, 0% loss) 192.168.2.1 : [1], 84 bytes, 0.37 ms (0.45 avg, 0% loss) 192.168.2.1 : [2], 84 bytes, 0.41 ms (0.44 avg, 0% loss)
````
First line is result, 
second, is the verbose output.

### CMD: `TEST=$(fping 192.168.2.1 -c3) ;echo; echo $TEST`

````bash

192.168.2.1 : 0.54 0.37 0.41

192.168.2.1 : [0], 84 bytes, 0.54 ms (0.54 avg, 0% loss) 192.168.2.1 : [1], 84 bytes, 0.37 ms (0.45 avg, 0% loss) 192.168.2.1 : [2], 84 bytes, 0.41 ms (0.44 avg, 0% loss)
````
First line is result, 
second, is the verbose output.

### CMD: `TEST=$(fping 192.168.2.1 -C3 -q) ;echo; echo $TEST`
Result no output
### CMD: `TEST=$(fping 192.168.2.1 -c3 -q) ;echo; echo $TEST`
Result no output

test on `fping: Version 4.1`

schweikert:
I agree that fping's use of stderr is not ideal, but we can't change it because of compatibility reasons (see also #9, which we had to revert).

PR #188, opened 2020-07-12
change all "fprintf(stderr," to "printf("

Eideen:
Hi
This PR will fix error in #184, by changing all "fprintf(stderr," to "printf("

### DIFF
--- a/src/fping.c
+++ b/src/fping.c
@@ -1311,7 +1311,7 @@ void main_loop()
 
 #if defined(DEBUG) || defined(_DEBUG)
             if (trace_flag) {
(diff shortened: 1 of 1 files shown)


schweikert:
I agree that fping's use of stderr is not ideal, but we can't change it because of compatibility reasons (see also #9, which we had to revert).

Eideen:
I think we need to break the compatibility, but we can allow a  period to allow migration to what is correct.

I think we can do the following:
1. Add a new flag for output to stdout (`--stdout` `--new`)and stderr (`--stderr` or `--old`), and add a warning to stdout that it will changed, in sometime. 
2. change the default to stdout.
3. delete the old  `--stderr`

Issue #199, opened 2020-08-24
fping: drops privileges too early for SO_BINDTODEVICE on Linux


hmh:
fping drops seteuid() privileges very early, before it processes the command line, and before it knows all setsockopt() it will need to do.

It does take care to create the IPv6-capable socket before dropping privileges, but this is not enough when SO_BINDTODEVICE (option -I <interface>) is required, at least on Linux 4.x and later.  setsockopt(SO_BINDTODEVICE) requires elevated privileges *at the time the setsockopt() is carried out*.

**This makes it impossible to _reliably_ use a setuid root fping to be used to ping IPv6 link-local destinations**.   It affects environments were something more elaborate (such as capabilities) are not available.  E.g.: OpenWrt.

The most immediate "fix" would be to drop seteuid privileges *after* all command line options are processed, and setsockopt() are issued.  This is likely the least troublesome way to address this, since for enhanced security, one should already be using some other mechanism (such as environment capabilities) instead of seteuid root.

hmh:
Note: it is possible to fping link-local destinations in IPv6, even in the FE80::/10 range, without SO_BINDTODEVICE.  But it won't be reliable if the neighbor table does not know that specific destination, or if the target (and maybe even source) addresses are assigned to several links/nodes -- something perfectly fine, since they are *link*-local.

hmh:
Pull request #200 is a proposal for a solution to this issue, using seteuid() to go back-and-forth between privileged and non-privileged effective UID.  Once we have issued all setsockopt that might need elevated privileges, drop them for good.

schweikert:
#200 has been merged.

PR #200, opened 2020-08-25
fping: retain privileges until after privileged setsockopt

hmh:
On Linux, one needs privileges to setsockopt(SO_BINDTODEVICE), and the
current code would drop setuid root privileges too soon.

Temporarily drop privileges instead, and raise them back to issue the
privileged setsockopt().  Once we know we won't need to do any further
privileged setsockopt(), permanently drop privileges.

For now, assume SO_BINDTODEVICE is the only setsockopt that needs this.

### DIFF
--- a/src/fping.c
+++ b/src/fping.c
@@ -397,6 +397,42 @@ void update_current_time();
 
 /************************************************************
 
+  Function: p_setsockopt
+
+*************************************************************
+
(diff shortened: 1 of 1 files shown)


hmh:
Related to issue #199

schweikert:
Looks good, thanks!

Issue #206, opened 2020-11-25
fping responses misdelivered on FreeBSD


clarsen:
On FreeBSD, the PIDs are as large as 99999, so if multiple fping processes are running there's a chance ident4 = PID & 0xffff will collide - which happens on monitoring hosts, like Nagios, that run fping often and simultaneously on a single IP address at a time.

If you have one IP up and one IP down and the two fping's PID ident4s collide, then the process pinging the down IP may see a reply from the up IP.

While the FreeBSD implementation of ping [calls connect()](https://github.com/freebsd/freebsd/blob/master/sbin/ping/ping.c#L644) which results in echo replies from undesired hosts being dropped before ever being delivered to the process, fping doesn't, so it has to rely on ICMP identifier to filter out the undesired responses.

It appears wait_for_reply() although it filters out ICMP identifier mismatch, it doesn't filter out on source IP address.

clarsen:
bump on this issue - there's a pull request we've been running with for 18 months or so so far...

eKrajnak:
I'm experiencing strange issue with fping 5.0 (bullseye repo) on Debian 11 (x64) and I think is colidate with this issue. Fping is used by Zabbix, we spawn 130 pinger processes simultaneously, usually busy for 40-50%, pinging 1k hosts every second (params -C10 -i0 -S [publicip]). Usually after VM reboot, random fping processes report random packet loss for random targets, even when I see icmp response in tcpdump. It's true that we ping a tens of permanently unavailable targets. This situation usually lasts few days, then disappear.

Is this bug affecting Linux distro also? Can we except fix for bullseye repos?
Thank you

hmh:
Once it is fixed upstream, you *have* to file a Debian bug explaining the situation, pointing to this bug, and identifying the affected (buggy) versions of fping and the fixed version.  *Then*, it depends on whether the downstream Debian fping maintainer will prepare a debian stable update for fping (possibly backporting only this fix).

gsnw:
@clarsen Do you happen to have an easy way to reproduce the error without using Nagios under FreeBSD?

auerswal:
I would expect `fping` to print out the differing IP address, e.g.:

```
$ fping 192.0.2.47
 [<- 192.0.2.33]192.0.2.47 is alive
```

This is not actual `fping` output, but based on the output of pinging a multicast address:

```
$ fping 224.0.0.1
 [<- 172.20.10.1]224.0.0.1 is alive
```

[...]

PR #207, opened 2020-11-25
Deal with multiple fping processes with colliding ICMP idents getting wrong echo reply

clarsen:
Fix for https://github.com/schweikert/fping/issues/206.
It's a slow linear search, but more simple is more correct.


schweikert:
I see multiple issues with this PR:
- The code is doing a linear search, so this could be impacting performance in a bad way
- fping does on purprose not make the assumption tha the source address of the reply matches the destination of the echo request

PR #211, opened 2021-01-25
Extract stats into their own file

TorunR:
Hi,

we've extracted some parts concerning connection stats into their own file during the course of a scientific study. We want to check with whether this extraction seems sensible for you.


schweikert:
I am not sure about the value of  doing this, and also this is creating conflicts at the moment, so I will close without merging.

Issue #216, opened 2021-03-07
Bogus values for -D if SO_TIMESTAMPNS is unavailable


sthen:
If the OS doesn't have SO_TIMESTAMPNS then fping chooses CLOCK_MONOTONIC for its timer if available. This results in bogus values for timestamps with fping -D as it uses an uptime-based clock rather than realtime.

```
$ fping -D -c1 127.0.0.1 2> /dev/null             
[596040.03537] 127.0.0.1 : [0], 64 bytes, 0.122 ms (0.122 avg, 0% loss)
```

This was introduced in 754a21e2bdf2 (https://github.com/schweikert/fping/blob/eac20347bac9aa396fa1c2039c57e7603862e6cc/src/fping.c#L127) when SO_TIMESTAMP was replaced with SO_TIMESTAMPNS.

sthen:
Oh, this is a duplicate of https://github.com/schweikert/fping/issues/203

Issue #203, opened 2020-09-14
Datetime with -Q behaves strangely


kaspergrubbe:
Hi,

I am running this program at `02:27` (at night/early morning) however the output from fping believes it is `14:42`, first I thought it interpreted `02:` as `14:`, but the minutes are also off by 15 minutes compared to the system clock:

```
$ echo $(date) && fping -B 1 -D -r0 -O 0 -p 1000 -Q 10 -l billetto.dk
Mon Sep 14 02:27:57 BST 2020
[14:42:09]
billetto.dk : xmt/rcv/%loss = 10/10/0%, min/avg/max = 22.6/24.8/33.1
[14:42:19]
billetto.dk : xmt/rcv/%loss = 10/10/0%, min/avg/max = 22.3/26.3/46.0
[14:42:29]
billetto.dk : xmt/rcv/%loss = 10/9/10%, min/avg/max = 22.2/24.5/31.6
```

# Version

Installed through Homebrew on OSX Catalina.

```
$ fping -v
fping: Version 5.0
```

gsnw:
It is reproducible on FreeBSD 12.2-RELEASE-p2, too but not on Debian 10.7

**FreeBSD** 
```
date +%s
1608499884

fping -D -l google.com
[877269.67467] google.com : [0], 64 bytes, 12.6 ms (12.6 avg, 0% loss)
[877270.70469] google.com : [1], 64 bytes, 12.4 ms (12.5 avg, 0% loss)
```
**Debian**
```
date +%s
1608500159

fping -D -l google.com
[1608500168.76708] google.com : [0], 64 bytes, 10.2 ms (10.2 avg, 0% loss)
[1608500169.75319] google.com : [1], 64 bytes, 9.92 ms (10.1 avg, 0% loss)
[1608500170.75338] google.com : [2], 64 bytes, 10.1 ms (10.1 avg, 0% loss)
```

gsnw:
At the build process the compiler don't set the macro  `#define CLOCKID CLOCK_REALTIME`
The following hot workaround fix temporary this issue on FreeBSD

```
diff --git a/src/fping.c b/src/fping.c
index c392fae..fb12eec 100644
--- a/src/fping.c
+++ b/src/fping.c
@@ -118,10 +118,11 @@ extern int h_errno;
 
 /*** Constants ***/
 
-#if HAVE_SO_TIMESTAMPNS
+//#if HAVE_SO_TIMESTAMPNS
[...]

PR #217, opened 2021-03-12
Fixed wrong timestamp under Free- and OpenBSD and macOS

gsnw:
Wrong timestamp on Free- and OpenBSD and macOS fixed by disabling #define CLOCKID CLOCK_MONOTONIC

**Issues**
#203
#216

### DIFF
--- a/src/fping.c
+++ b/src/fping.c
@@ -122,8 +122,11 @@ extern int h_errno;
 #define CLOCKID CLOCK_REALTIME
 #endif
 
+/* CLOCK_MONTONIC starts under macOS, OpenBSD and FreeBSD with undefined positive point and can not be use
+ * see github PR #217
+ */
 #if !defined(CLOCKID)
-#if defined(CLOCK_MONOTONIC)
+#if defined(CLOCK_MONOTONIC) && !defined(__APPLE__) && !defined(__OpenBSD__) && !defined(__FreeBSD__)
 #define CLOCKID CLOCK_MONOTONIC
 #else
 #define CLOCKID CLOCK_REALTIME


tohojo:
Using CLOCK_MONOTONIC breaks the timestamp output just as much under Linux as under BSD (but this is hidden by HAVE_SO_TIMESTAMPNS being set so the clock is always CLOCK_REALTIME). So why is CLOCK_MONOTONIC there at all?

hmh:
Hmm, you really ought to use CLOCK_MONOTONIC for every time-delta calculation (unless you're using packet timestamps from either kernel or hardware, which is highly OS-specific).  Severe weirdness happens when CLOCK_REALTIME steps around due to system administrator action (or even non-slew-mode large ntp corrections, etc).

If you need human-readable timing, you timestamp CLOCK_REALTIME and CLOCK_MONOTONIC at program start (or at measurement session start, if there's such a notion in your application -- for fping they're the same), and use the CLOCK_MONOTONIC delta + CLOCK_REALTIME timestamp to calculate a more sane real time that doesn't jump around.

This adds a minor, constant, and irrelevant phase error to the real time output, which is already quite uncertain anyway.

src/fping.c
```diff
@@ -123,7 +123,7 @@ extern int h_errno;
 #endif
 
 #if !defined(CLOCKID)
-#if defined(CLOCK_MONOTONIC)
+#if defined(CLOCK_MONOTONIC) && !defined(__APPLE__) && !defined(__OpenBSD__) && !defined(__FreeBSD__)
```
schweikert, line 126: Could you add a comment to explain why we can't use CLOCK_MONOTONIC on these platforms?

gsnw: On FreeBSD and OpenBSD CLOCK_MONOTONIC begin with 0 after reboot or undefined positive point.
Here is my test result with FreeBSD.
```
CLOCK_MONOTONIC (sec): 43
CLOCK_REALTIME (sec): 1628355889
CLOCK_UPTIME (sec): 43
```
OpenBSD Manpage: https://man.openbsd.org/clock_gettime.2

macOS unfortunately I can not test only assume

schweikert: Thanks, I meant also a comment in the code. Maybe referencing just this PR, something like this:

// don't use CLOCK_MONOTONIC on Mac and FreeBSD because gettime returns uptime (see github PR #217)

gsnw: Ups, sorry for misunderstanding. I have add a in code comment

schweikert: Thanks!

Issue #171, opened 2020-02-12
[Feature Request] Stop pinging if N hosts are reachable


mirfatif:
In combination with `--reachable=N` it would save time on large networks. E.g. with `fping --retry=5 --timeout=200 --reachable=2 --generate 192.168.1.0/24` if router responds at `192.168.1.1` and any other host at DHCP assigned IP - say `192.168.1.10` - do not retry and wait for timeout on other hosts.

gsnw:
I have created a new pull request #260 with complete copy of the code and a corresponding test case

mirfatif:
I'm not currently using `fping`. If the functionality has been added, please close.

PR #220, opened 2021-05-24
fping: add option to exit immediately once N hosts have been found

chriscray:
This is a feature that we'd really like to have, and given #171 there seems to be some interest in the larger community as well.


schweikert:
This looks good, thank you (and sorry for the late answer). Can you maybe add a test for this?

schweikert:
Replaced by #260 which is based on this and added a test case. Thanks!

Issue #223, opened 2021-06-29
Travis CI - switch to GitHub Actions?


darless:
The building on travis-ci.org has ceased. I see that the build passing for this repo still points to org version. The com version has limits on opensource. If there is interest in using Github Actions - I could look into migrating the travis configuration to github action workflows.

darless:
Closing this as GH functionality was added since the time this was opened. 

```
commit 3ecfbb7c1e1743c6cda3bf270a88eedcb2e292f8
Author: David Schweikert <david@schweikert.ch>
Date:   Sun Jan 23 18:04:19 2022 +0100

    GH action test
```

PR #224, opened 2021-06-30
Dockerfile for fping

darless:
This adds a Dockerfile to the repository. The start of this is to build on the latest ubuntu (20.04).

I could see this being useful to test out new changes that require different versions of packages without affecting the host system. This is also useful if/when CI is modified to use Github Actions #223 as the dockerfile could automatically get pushed to the latest on a push to master. 

Build:
```
docker build -t fping -f contrib/Dockerfile .
```

Run fping on a host that does not have fping currently
```
$ docker run -it fping 4.2.2.2 8.8.8.8 1.2.3.4
4.2.2.2 is alive
8.8.8.8 is alive
1.2.3.4 is unreachable
```

Run fping within container
```
docker run --entrypoint bash -it fping
root@8ca8ee184d27:/app# fping 4.2.2.2 1.2.3.4 -t 30
4.2.2.2 is alive
1.2.3.4 is unreachable
```


schweikert:
Thank you, I am willing to merge this, but could you move it under "contrib"?

darless:
This has been done and updated original comment about how to build with Dockerfile in the contrib folder.

darless:
@schweikert what else is needed for PR to be merged?

schweikert:
I still see a file named Dockerfile outside of the contrib directory.

darless:
You are correct, my apologies, should be corrected now. I left .dockerignore in the main directory, to allow building appropriately. If that's not accepted I can figure out how best to change the building of the docker container.

schweikert:
Thanks!

PR #227, opened 2021-08-07
fix getnameinfo

aafbsd:
getnameinfo() was not called properly for IPv4 (didn't check v6).
Additionally, one might also want to check the result code, call gai_strerror() and inform the user if s.th. goes wrong


schweikert:
Thanks!

Issue #218, opened 2021-04-27
PATCH: fix getnameinfo


aafbsd:
getnameinfo() is not called properly for IPv4 (didn't check v6).

[patch-d-fix_getnameinfo.txt](https://github.com/schweikert/fping/files/6382239/patch-d-fix_getnameinfo.txt)

You might also want to check the result code, call gai_strerror() and inform the user if s.th. goes wrong

schweikert:
Could you create a pull request?

aafbsd:
I am no real git or github user so this is a first wild guess. Whatever I messed up should be fixable. Constructive criticism is always welcome :-).

gsnw:
Pull request: #227

gsnw:
@schweikert you can close this issue. That fixed with pull request #227

PR #230, opened 2021-08-20
Fix getnameinfo() is not called properly for IPv4

gsnw:
Add Patch #218 from @aafbsd to pull request.


gsnw:
Sorry my mistake there is already a pull-request #227

PR #245, opened 2022-02-05
add missing files to autoclean.sh

auerswal:
At least on my system (Ubuntu GNU/Linux 18.04.6 LTS), autoclean.sh
misses three files: compile, doc/Makefile, and src/Makefile.  Add
them.


auerswal:
I think that the *coverage/coveralls* result is bogus since this pull request touches neither `fping` code nor tests.

PR #249, opened 2022-02-13
fix unprivileged mode on Linux

schweikert:


src/fping.c
```diff
@@ -1104,11 +1104,11 @@ int main(int argc, char** argv)
         exit(num_noaddress ? 2 : 1);
     }
 
-    if (src_addr_set && socket4 >= 0) {
+    if (socket4 && (src_addr_set || socktype4 == SOCK_DGRAM)) {
```
sfan5, line 1107: Shouldn't this still be `socket4 >= 0`? 0 is theoretically a valid fd while -1 is not

schweikert: Yes, thanks for catching this.

src/fping.c
```diff
@@ -1104,11 +1104,11 @@ int main(int argc, char** argv)
         exit(num_noaddress ? 2 : 1);
     }
 
-    if (src_addr_set && socket4 >= 0) {
+    if (socket4 >= 0 && (src_addr_set || socktype4 == SOCK_DGRAM)) {
         socket_set_src_addr_ipv4(socket4, &src_addr, (socktype4 == SOCK_DGRAM) ? &ident4 : NULL);
     }
 #ifdef IPV6
-    if (src_addr6_set && socket6 >= 0) {
+    if (socket6 >= 0 && (src_addr6_set || socktype4 == SOCK_DGRAM)) {
```
sfan5, line 1111: One more: it says `socktype4` on this line, copy-paste mistake?

schweikert: Oh my... fixed as well. thanks!

Issue #201, opened 2020-09-03
fping not working on Synology DiskStation.


bolemo:
Tried from Entware, and compiling myself.

It installs, I can change permissions for setuid, but only fping -h returns something. Anything else returns nothing (and return code 4), even fping -v

```
root@Mnemosyne:~# ls -l /usr/local/sbin/fping 
-rwsr-xr-x 1 root root 115240 Sep  3 18:19 /usr/local/sbin/fping
root@Mnemosyne:~# file /usr/local/sbin/fping 
/usr/local/sbin/fping: setuid ELF 32-bit LSB executable, ARM, EABI5 version 1 (SYSV), dynamically linked, interpreter /opt/lib/ld-linux.so.3, for GNU/Linux 2.6.32, with debug_info, not stripped
root@Mnemosyne:~# fping -v
root@Mnemosyne:~# echo $?
4
root@Mnemosyne:~# uname -a
Linux Mnemosyne 3.2.40 #25426 SMP Wed Jul 8 03:10:10 CST 2020 armv7l GNU/Linux synology_comcerto2k_ds414j
```

Kiskae:
I've run into this issue while running a newer version of the `debian` docker container. After looking at the system calls it appears to be caused by the absence of the `/etc/protocols` file causing the `getprotobyname` library call to fail.

While the code does include a message for this problem the error happens before the options are parsed and `verbose_flag` can be set, so the message never gets printed to `stderr`.

Considering `crash_and_burn` only happens in during an error it might be a good idea to remove the `verbose_flag` condition and always print the error. @schweikert

georg-nielsen:
Hi,
Don't know if you have already solved this issue but I ran into the same and the solution was following when building fping from source.  Disable IPv6 and enable IPv4:
./configure --disable-ipv6 --enable-ipv4 --prefix=/opt
and then make and  make install

Best regards,
Georg.

schweikert:
We should implement a better failure mode when this happens, i.e. output an error when getprotobyname fails. Anybody willing to create a pull request?

shaferandrew:
I was experiencing this issue on an ASUSTOR AS6604 NAS with fping 5.0-1 installed via entware.

I fixed it by copying the /etc/protocols file from an Ubuntu 21.10 machine to the NAS /etc/protocols file.  The Ubuntu file was significantly longer and had more entries, specifically for IPv6 protocols.

fping responded correctly after that.

auerswal:
Hi,

[...]

PR #252, opened 2022-03-26
emit error message when socket creation fails

auerswal:
This is the patch from my [comment on issue 201](https://github.com/schweikert/fping/issues/201#issuecomment-1077475551), but now I have tested it, and it works, at least for me. :-)

To test, I have temporarily changed the requested protocol name to something nonsensical. With the patch, an error message was emitted, but not without the patch.

I think a pull request like this was requested in a [comment on issue 201](https://github.com/schweikert/fping/issues/201#issuecomment-894614950).

My PR 250 (distinguish ICMP and ICMP6 in error messages) makes this error message more useful.

### DIFF
--- a/src/fping.c
+++ b/src/fping.c
@@ -463,6 +463,9 @@ int main(int argc, char** argv)
         usage(0);
     }
 
+    /* enable error messages during socket creation */
+    verbose_flag = 1;
+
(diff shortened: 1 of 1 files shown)


auerswal:

PR #254, opened 2022-05-01
Ipv6 generator

auerswal:
IPv6 support for -g, --generate with open-coded 128-bit arithmetic as needed.

### DIFF
--- a/ci/test-06-options-f-h.pl
+++ b/ci/test-06-options-f-h.pl
@@ -1,6 +1,6 @@
 #!/usr/bin/perl -w
 
-use Test::Command tests => 45;
+use Test::Command tests => 48;
 use File::Temp;
 
 #  -f file    read list of targets from a file ( - means stdin) (only if no -g specified)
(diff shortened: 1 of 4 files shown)


auerswal:
Coveralls on macos reports reduced coverage, because it looks at the code *with* IPv6 generator support (i.e., `-DIPV6`), but the *tests for this new code are skipped*, because the tests on macOS are run with `SKIP_IPV6`.

One IPv6 test on the Ubuntu based test systems fails, because they do not have an IPv6 default route, and Linux only installs `fe80::/64` routes, not `fe80::/10` routes, when an interface has a link-local IPv6 address.  One of the new tests sends to address `fe80:0:0:1::` (to test crossing the 64-bit boundary during address generation) which is just outside of `fe80::/64`. I did not see this on my development system, because my system does have an IPv6 default route.

When adding a `blackhole`, `unreachable`, or `prohibit` route for `fe80::/10` to my development system, `fping` reports errors (`Invalid argument` for `blackhole`, `No route to host` for `unreachable`, and `Permission denied` for `prohibit`).  Thus this would not change the test outcome.

I see at least two methods to address this problem:

1. I could remove the test for an empty `stderr` for this test.
2. I could add another `SKIP_…` variable, e.g., `SKIP_IPV6_DEFAULT_ROUTE`.

I think it is OK to not look at `stderr` for this test, because `stdout` is tested to contain all addresses that shall be generated, and the test primarily checks if *generating* addresses works, not if sending packets to the generated addresses works, too.  Additionally, other tests do verify an empty `stderr` when sending to generated link-local addresses.

I'll add a patch to change this test to ignore `stderr` to this pull request soon.

auerswal:
The **coverage/coveralls** result seems bogus. It reports that test coverage of the files `seqmap.c`, `socket4.c`, and `socket6.c` has changed, but I did not remove any existing tests, and did not change those files.

The **Coveralls - macos** result is caused by building with IPv6 support on macOS, but not testing it. But building without IPv6 support and skipping IPv6 tests results in existing tests in file `test-04-options-a-b.pl` to fail, so this does not seem to be a good solution (without adjusting those tests).

auerswal:

xtaran:
Please reopen.

Issue #262, opened 2022-08-19
fping -Q does not output to standard output


ziqingkeshi:
Hello,
Thank you very much for the program.
But I have encountered some problems with using fping,here is my system info:

**Linux version 3.10.0-1160.el7.x86_64 (mockbuild@kbuilder.bsys.centos.org) (gcc version 4.8.5 20150623 (Red Hat 4.8.5-44) (GCC) ) #1 SMP Mon Oct 19 16:18:59 UTC 2020**

When I want to redirect the result of `fping -Q` command to a file, it didn't work, and output the result to the console,and there is no result in the file either.

![image](https://user-images.githubusercontent.com/41533033/185530509-b57a5d0a-e35b-404c-9477-05a49c9732fb.png)

I tried all kinds of methods, but they didn't work.Finally, I achieved by redirecting stderr:

![image](https://user-images.githubusercontent.com/41533033/185531957-c8d6e24f-fe46-44eb-8520-c5d1b74a62fc.png)

It looks like when "**-Q**" is used, the output results to standard error. This problem does not occur when the "**-Q**" parameter is not added, and the results can be obtained from standard output or redirected to a file normally.

I'm not sure if this is an issue and I've been able to achieve what I need, but I still want to figure out this phenomenon.

Looking forward to your reply.
Thank you very much！

darless:
The `-Q` flag as you say uses stderr. This calls the `print_per_system_splits`. Using the `-N` flag uses `print_netdata`. With `-N` flag, stdout redirection works without having to redirect stderr. I do not know why it is important to use stderr with `print_per_system_splits`. I created a PR for this to change to stdout.

schweikert:
While I agree that fping's use of stdout and stderr is not consistent and could be better, we can't change it now because it would break users calling fping programmatically (i.e. it changes its interface).

PR #269, opened 2022-12-06
Address #262: Update print_per_system_splits to use stdout instead of…

darless:
There is an inconsistency in how the output is printed.

'-Q' - report_interval - prints to stderr
'-N' - netdata_flag - prints to stdout.
'-s' - stats_flags - prints to stderr

With this change, the '-Q' will also print to stdout instead of stderr. If there is a legitimate reason why there is an inconsistency, then please specify it in the documentation.

### DIFF
--- a/src/fping.c
+++ b/src/fping.c
@@ -1730,40 +1730,40 @@ void print_per_system_splits(void)
     struct tm *curr_tm;
 
     if (verbose_flag || per_recv_flag)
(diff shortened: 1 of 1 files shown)


schweikert:
We can't change this because it is part of the API of fping and many programs using fping would break. I agree that it is not ideal, but that's just the way it is.\

Issue #273, opened 2023-04-11
setting SO_TIMESTAMPNS option: Protocol not available (on debian running in docker container)


rolf-d2i:
When running fping 5.1 in a docker container on MacOSX Ventura 13.0 in emulated amd64 mode on apple silicon I got 

>>fping
setting SO_TIMESTAMPNS option: Protocol not available
setting SO_TIMESTAMPNS option (IPv6): Protocol not available

The error can be fixed by downgrading fping to for instance the 4.2 version by downloaded fping package from debian package site and installing the old package using dpkg. 
Running fping 5.0 in a native  debian docker container on apple silicon works as expected (currently latest version). 
Both the arm and amd64 debian versions were 11.6.

gsnw:
I think it is the same problem as #272

rolf-d2i:
It may be the same issue but the previous error was specific to windows server and WSL1 and is therefore Windows specific problem. Issue #272 should perhaps be rewritten to include every instance of operating system machine combination that has encountered the same issue?

gsnw:
SO_TIMESTAMPNS exists since fping 5.0
fping 4.3 works because it is not present there.

When building the lates version it is checked if "SO_TIMESTAMPNS" is set on the system.
The package builders of the distributions apparently use a kernel where this is present.
If now this package is installed on a system e.g. amd64, where a different kernel is present, which does not support "SO_TIMESTAMPNS", then it should come to this message.

This is probably the same for Microsoft with WSL.

rolf-d2i:
The amd64 emulation on MacOSX lacks support for some vector instructions on amd64/intel cpus and possible is missing some other cpu instructions. The kernal is supposed to be a standard debian linux kernal that can run on any amd64 processor., I doubt the issue here is the kernal itself.

gsnw:
Can you build https://github.com/gsnw/fping/tree/issue/272-273 and check this?
I have add a runtime check for SO_TIMESTAMPNS with fallback to SO_TIMESTAMP

I hope this help and works.

rolf-d2i:
After building
root@f74a2b206ad3:~/fping# /usr/local/sbin/fping -v
/usr/local/sbin/fping: Version 5.1
root@f74a2b206ad3:~/fping# /usr/local/sbin/fping www.google.se
setting SO_TIMESTAMPNS option: Protocol not available
setting SO_TIMESTAMPNS option (IPv6): Protocol not available
www.google.se is alive

Still the same issue

[...]

PR #279, opened 2023-08-22
Fallback to SO_TIMESTAMP if SO_TIMESTAMPNS is not available

gsnw:
If the compiler correctly identifies the use of HAVE_SO_TIMESTAMPNS but in runtime SO_TIMESTAMPNS cannot be used, then SO_TIMESTAMP is tried to be used as fallback. Only if this also does not work, an error message is issued for both SO_TIMESTAMPNS and SO_TIMESTAMP.

Fixed Issues #272 and #273


kaspergrubbe:
Thank you! :)

PR #289, opened 2023-12-15
Add --fwmark option for Linux

deepkv:
This is based on [tomangert](https://github.com/tomangert)'s original outdated pull request.

--fwmark can be used in conjunction with `ip rule` and `ip route` to route ping packets based on the mark set by the aforementioned flag.

This has been supported by Linux since kernel version [2.6.25](https://man7.org/linux/man-pages/man7/socket.7.html)

Let me know if you'd like any changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,13 +28,14 @@ If you want to install fping from source, proceed as follows:
    (see: `./configure --help`)
 2. Run `make; make install`.
 3. Make fping either setuid, or, if under Linux:
-   `sudo setcap cap_net_raw+ep fping`
+   `sudo setcap cap_net_raw,cap_net_admin+ep fping`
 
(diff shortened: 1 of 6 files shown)


schweikert:
Thanks for your contribution!

Could you add a simple test to check at least that the program doesn't crash when --fwmark is passed? Pinging 12.0.0.7 is fine, and you don't need to check that the fwmark was actually set (that would be hard to do).

Also, please update doc/fping.pod (the man-page).

deepkv:
Thanks for the review! Done.

deepkv:
Apologies, it looks like setcap doesn't work in GitHub CI, which made the test fail. I've now used sudo instead. I believe it should work now.

deepkv:
Update: the tests finally pass in GitHub CI in my repo after a couple more iterations. Thanks for your patience.

Issue #288, opened 2023-12-12
not enough sequence numbers available! (expire_timeout=10000000000, host_nr=0, ping_count=0, seqmap_next_id=0)


cagney:
On FreeBSD (14.0, but also 13.x), fping gets this error:
```
+# fping  -c 1  --timeout 1s   --src 192.0.1.254 192.0.2.254
+fping error: not enough sequence numbers available! (expire_timeout=10000000000, host_nr=0, ping_count=0, seqmap_next_id=0)
```
when it is run immediately after a boot (and I really mean immediately, its from a test framework).   Adding a sleep 10 before running fping seems to fix the problem.

I suspect:
```
    /* check if expired (note that unused seqmap values will have fields set to
     * 0, so will be seen as expired */
    next_value = &seqmap_map[seqmap_next_id];
    if (timestamp - next_value->ping_ts < SEQMAP_TIMEOUT_IN_NS) {
        fprintf(stderr, "fping error: not enough sequence numbers available! (expire_timeout=%" PRId64 ", host_nr=%d, ping_count=%d, seqmap_next_id=%d)\n",
            SEQMAP_TIMEOUT_IN_NS, host_nr, ping_count, seqmap_next_id);
        exit(4);
    }
```
where, immediately after a boot `timestamp` is small which means `timestamp-0/*ping_ts*/` is less than `SEQMAP_TIMEOUT_IN_NS` i.e., 10s if my math is correct.  The fix would be to set `.ping_ts` to some equivalent of the epoc.

I also suspect  #217 was wrong.

gsnw:
Unfortunately I cannot reproduce the error, but I think it is related to the jumping time mentioned by hmh in the pull-request, which can occur with CLOCK_REALTIME.
Presumably NTP is used on the system?

The patch is unfortunately not very clean in this respect, but should primarily correct the time output, which it does.
Unfortunately, it seems to generate unwanted subsequent errors in certain constellations.

I'll take a look at the whole thing and look for a better solution. HMH has already mentioned an approach here

> and use the CLOCK_MONOTONIC delta + CLOCK_REALTIME timestamp to calculate a more sane real time that doesn't jump around.

Whereby the time output also becomes less accurate at some point with a longer runtime of the system, but the measurement result is not falsified.

gsnw:
Can you please check whether the problem is also present with the version under a3d991b ?
Of course, this is not the final version. I just want to know whether the error is gone.

cagney:
[...]

PR #290, opened 2023-12-19
Use CLOCK_REALTIME when logging and CLOCK_MONOTONIC when timing

cagney:
Change update_current_time() so that current_time is set from CLOCK_REALTIME and current_time_ms is set from CLOCK_MONOTONIC.  This way log messages display wall clock time while any timing calculations use monotonic time.

See #203

In seqmap_add(), before checking to see if .time_ts wrapped, check that .time_ts was set.

See #288

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -9,19 +9,6 @@ m4_ifdef([AC_AUTOCONF_VERSION],[AC_USE_SYSTEM_EXTENSIONS], [AC_GNU_SOURCE])
 
 # Detect Operatingsystem
 AC_CANONICAL_TARGET
-only_clock_realtime=no
-
-case "${target}" in
-  *darwin*)
-    only_clock_realtime=yes
-    ;;
-  *freebsd*)
-    only_clock_realtime=yes
-    ;;
-  *openbsd*)
-    only_clock_realtime=yes
-    ;;
-esac
 
 dnl --disable-ipv4
 AC_ARG_ENABLE([ipv4],
@@ -60,7 +47,6 @@ dnl Test if --enable-timestamp is explicitely enabled and make an error if this 
 AS_IF([test "x$enable_timestamp" = "xyes" -a "x$have_so_timestamp" = "xno"], [
   AC_MSG_ERROR([--enable-timestamp not supported on this platform])
 ])
-AS_IF([test "x$only_clock_realtime" = "xyes"], [AC_DEFINE(ONLY_CLOCK_REALTIME, [1], [ONLY_CLOCK_REALTIME is defined])])
 
 AC_ARG_ENABLE([safe-limits],
   AS_HELP_STRING([--enable-safe-limits], [Restrict timing parameters (-i, -p) within "safe" limits]))
--- a/src/fping.c
+++ b/src/fping.c
@@ -114,24 +114,6 @@ extern int h_errno;
 }
 #endif /* __cplusplus */
 
-/*** Constants ***/
-
-/* CLOCK_MONTONIC starts under macOS, OpenBSD and FreeBSD with undefined positive point and can not be use
- * see github PR #217
- * The configure script detect the predefined operating systems an set CLOCK_REALTIME using over ONLY_CLOCK_REALTIME variable
- */
-#if HAVE_SO_TIMESTAMPNS || ONLY_CLOCK_REALTIME
-#define CLOCKID CLOCK_REALTIME
-#endif
-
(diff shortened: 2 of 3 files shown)


schweikert:
This looks good, thank you! Could you have a look at the failing test and check whether the failure is caused by this change?

cagney:
looking

gsnw:
The test is broken by fping.c line 1500 `current_time_ns = timespec_ns(&monotonic_time);`
When I turn it back to CLOCK_REALTIME `current_time_ns = timespec_ns(&current_time);`, the test work

I hope this helps with the analysis

cagney:
Off hand, would you know which clock:
```
    /* ancilliary data */
    {
        struct timespec reply_timestamp_ts;
        for (cmsg = CMSG_FIRSTHDR(&recv_msghdr);
             cmsg != NULL;
             cmsg = CMSG_NXTHDR(&recv_msghdr, cmsg)) {
            if (cmsg->cmsg_level == SOL_SOCKET && cmsg->cmsg_type == SCM_TIMESTAMPNS) {
                memcpy(&reply_timestamp_ts, CMSG_DATA(cmsg), sizeof(reply_timestamp_ts));
                *reply_timestamp = timespec_ns(&reply_timestamp_ts);
            }
        }
    }
```
the Linux kernel uses for this?

cagney:
I'm pretty sure it is CLOCK_REALTIME, which, on linux, leads to:
```
$ ./src/fping -z 0xffff 127.0.0.1
./src/fping:
  count: 1, retry: 3, interval: 10 ms
  perhost_interval: 1000 ms, timeout: 500
  ping_data_size = 56, trials = 4
  verbose_flag set
  backoff_flag set
  report_all_rtts_flag set
  randomly_lose_flag set
  print_per_system_flag set
[83452.72502] 127.0.0.1 [0]: add ping event in 0 ms
[83452.72502] # main_loop
[83452.72502] 127.0.0.1 [0]: ping event
[83452.72505] seqmap_add(host: 0, index: 0) -> 0
[83452.72505] 127.0.0.1 [0]: send ping
[83452.72505] 127.0.0.1 [0]: add timeout event in 500 ms
[83452.72505] next timeout event in 500 ms (127.0.0.1)
[83452.72505] waiting up to 500 ms
[83452.72822] seqmap_fetch(0) -> host: 0, index: 0
[83452.72822] received [0] from 127.0.0.1 recv_time=1703084424276482893 (curr_time=83452728221621) this_reply=1703000971551429542 timeout=500000000
[83452.72822] discard [0] from 127.0.0.1 as this_reply=1703000971551429542 exceeds timeout=500000000
```
it's looking like a better option is to just and only use CLOCK_REALTIME and live with delta changes.

gsnw:
Unconfirmed and unknowingly, I would say yes.
socket.c seems to use `ktime_get_real()` and that again refers to CLOCK_REALTIME

https://www.kernel.org/doc/html/latest/core-api/timekeeping.html
https://github.com/torvalds/linux/blob/master/net/core/sock.c

gsnw:
Your patch seems to work on FreeBSD.
If we limit this to FreeBSD, OpenBSD and macOS as before, we have a solution.

cagney:
My thought, having worked through this, is to reject this patch and instead simplify the conditional code so it always uses CLOCK_REALTIME.  The real cause of the FreeBSD problems is their out-of-date package.  I'll poke them.

The ideal clock source is probably CLOCK_BOOTTIME but that's linux only and won't work with timestamp.

gsnw:
I may have found an approach.
If we swap select for epoll, CLOCK_MONOTONIC might work under Linux.

https://man7.org/linux/man-pages/man2/epoll_wait.2.html

gsnw:
I have found a solution that fixes your problem with the test on Linux 1a83be3fce5f29839ccf79d33d0fa289cac225e2
But they don't work under macOS and probably not under BSD either, because #include <linux/net_tstamp.h> doesn't exist there.
So still needs to be refined.

rowingdude:
Is there still ambition of improving this branch? I am looking for a project.

gsnw:
@rowingdude you are free to look for a final solution. Currently, only a part without CLOCK_MONOTONIC is used as the primary time specification. You can see the reason for this in the previous posts.

Otherwise I would close the pull request at the end of the month.

gsnw:
Closed because the pull request is not functional. There is also no corrected version.
Parts of the pull-request have been transferred via a new pull-request.

PR #291, opened 2023-12-21
top view for fping

lordrasmus:
i made some changes to have another option to view the pings

i thik the code needs a bit cleanup and fixes but i wanted to ask if you whould like that

it looks like this
ping


schweikert:
It looks nice and thanks for your contribution, but I would like to keep the number of "non-core" features in fping to a minimum. It should be possible to write a command-line tool that does the UI and calls fping for the measurements.

lordrasmus:
I believe crafting a tool solely invoking fping isn't as straightforward, as it needs to do more than just display the results.

It should run continuously and provide real-time status updates.

This might pose a challenge, as parsing the output as text would be necessary.

PR #294, opened 2024-01-10
Update exit status description in man page

auerswal:
With the options -x and -X, fping does not require all hosts to be reachable for an exit status of 0.

This addresses a comment in issue #99 suggesting such a man page update.


auerswal:
The coverage decrease is obviously bogus, because this pull request changes only the POD file, but neither code nor tests. Please consider merging despite the unsuccessful check. Thanks!

auerswal:
Thanks!

Issue #295, opened 2024-02-04
Combinations "-c -a" and "-c -u" give same output as combination "-c -q" instead of just "-c"


auerswal:
Both options `-a` (`--alive`) and `-u` (`--unreachable`) should probably just be ignored with `-c N` (`--count N`), but they seem to have the same effect as `-q` (`--quiet`).

`fping` version:

```
$ fping -v
fping: Version 5.1
```

Expected behavior:

```
$ fping -c3 8.8.8.{7,8}
8.8.8.8 : [0], 64 bytes, 74.9 ms (74.9 avg, 0% loss)
8.8.8.7 : [0], timed out (NaN avg, 100% loss)
8.8.8.8 : [1], 64 bytes, 63.8 ms (69.3 avg, 0% loss)
8.8.8.7 : [1], timed out (NaN avg, 100% loss)
8.8.8.8 : [2], 64 bytes, 73.8 ms (70.8 avg, 0% loss)
8.8.8.7 : [2], timed out (NaN avg, 100% loss)

8.8.8.7 : xmt/rcv/%loss = 3/0/100%
8.8.8.8 : xmt/rcv/%loss = 3/3/0%, min/avg/max = 63.8/70.8/74.9
```
```
$ fping -c3 -q 8.8.8.{7,8}
8.8.8.7 : xmt/rcv/%loss = 3/0/100%
8.8.8.8 : xmt/rcv/%loss = 3/3/0%, min/avg/max = 60.8/61.7/62.4
```

Unexpected behavior:

```
$ fping -c3 -a 8.8.8.{7,8}
8.8.8.7 : xmt/rcv/%loss = 3/0/100%
8.8.8.8 : xmt/rcv/%loss = 3/3/0%, min/avg/max = 72.8/76.3/82.1
```
```
$ fping -c3 -u 8.8.8.{7,8}
8.8.8.7 : xmt/rcv/%loss = 3/0/100%
8.8.8.8 : xmt/rcv/%loss = 3/3/0%, min/avg/max = 50.1/55.6/65.2
```

(I have noticed this while looking into issues #271 and #274.)

Issue #274, opened 2023-04-28
-u does not work when -c is used


asarubbo:
```
 ~ $ fping -v
fping: Version 5.1
```

This works for me:
```
~ $ fping -4 -r 1 -t 1000 10.10.5.90 10.10.5.1 -q -u
10.10.5.90
```

However when I add `-c` it doesn't
```
 ~ $ fping -4 -r 1 -t 1000 10.10.5.90 10.10.5.1 -q -u -c 1
10.10.5.90 : xmt/rcv/%loss = 1/0/100%
10.10.5.1  : xmt/rcv/%loss = 1/1/0%, min/avg/max = 0.171/0.171/0.171
```
Because it also prints about reachable targets.

auerswal:
I am not really sure what you want to achieve with the above `fping` invocation.

It seems to me as if `fping -u -r 1 10.10.5.90 10.10.5.1` already provides all the interesting information. So what purpose have the options `-c 1 -q`? Do you attempt to get the following output (I am using two different IP addresses for illustration)?

```
$ fping -r1 -q -u -c1 8.8.8.7 8.8.8.8 2>&1 | grep -v ,
8.8.8.7 : xmt/rcv/%loss = 1/0/100%
```
Since the `fping` command line specifies to send just one echo request, the statistics can be added with, e.g., `sed`:

```
$ fping -r1 -u 8.8.8.7 8.8.8.8 | sed 's,$, : xmt/rcv/%loss = 1/0/100%,'
8.8.8.7 : xmt/rcv/%loss = 1/0/100%
```

It seems to me as if just `-u -r N` already provides all the information the combination of `-r 1 -c N -q -u` can give.

asarubbo:
> I am not really sure what you want to achieve with the above `fping` invocation.

I was just saying that by adding `-c 1`  (so just a count) it prints out the reachable targets and their stats, while with `-q` it shouldn't.

auerswal:
The option `-c N` prints one output line for every received reply or timeout, and one summary line per target at the end.

The combination `-c N -q` only prints the summary line per target, but not the individual per-probe reports.

If you want the output format of `-u`, i.e., just the Name or IP address of each unreachable target, but want `fping` to try several times to reach each target, you can use `-r N` optionally combined with `-B 1`, instead of using `-c N`. (By default, `fping` uses `-r 3`.)

Issue #271, opened 2023-02-27
--alive and --count options mutual exclusive


psuet:
When using the --alive option in conjunction with the --count option, the --alive option is simply disregarded without any error or warning.

This behaviour is unintuitive, and the mutual exclusivity is not documented as far as I can see.

auerswal:
Do you expect only the lines for targets for which no response at all was received to be omitted? E.g., as follows?

```
$ ./src/fping -u -c1 8.8.8.7 8.8.8.8 2>&1 | grep -v '/0/100%'
8.8.8.8 : xmt/rcv/%loss = 1/1/0%, min/avg/max = 80.5/80.5/80.5
```

psuet:
Yes, similar, but --alive only outputs the IP instead of all this information.

```
$ ./src/fping -u -c1 8.8.8.7 8.8.8.8 2>&1 | grep -v '/0/100%' | cut -d: -f1 | xargs
8.8.8.8
```

My (former) usecase:
I have a large Layer 2 Network (i know, i know) with some specialized hardware that sometimes swallows / rate-limits requests on first try. Thus I have to repeat requests to really get an accurate list of all alive devices.

auerswal:
Ah, I made a mistake in my example, it should not have had the `-u`. I was thinking about:

```
$ fping -c3 8.8.8.7 8.8.8.8 2>&1 | grep -v -e '/0/100%' -e 'timed out'
8.8.8.8 : [0], 64 bytes, 65.3 ms (65.3 avg, 0% loss)
8.8.8.8 : [1], 64 bytes, 61.3 ms (63.3 avg, 0% loss)
8.8.8.8 : [2], 64 bytes, 64.2 ms (63.6 avg, 0% loss)

8.8.8.8 : xmt/rcv/%loss = 3/3/0%, min/avg/max = 61.3/63.6/65.3
```

I would expect that the `--retry=N` option together with `--backoff 1` should help with your use case:

```
$ fping --retry 10 --backoff 1 --alive 8.8.8.7 8.8.8.8 
8.8.8.8
```

With `--backoff 1`, the pause between attempts is constant, as with `--count N`, but `fping` stops sending echo requests once it has received an answer.

PR #296, opened 2024-02-05
Options "-c N" / "-C N" versus "-a" and "-u"

auerswal:
1. Adjust `fping -h` output to emphasize that `-c N` changes the output format.
2. Adjust `fping -h` output to emphasize that `-i MSEC` does not change the output format.
3. Ensure that neither `-a` nor `-u` affect `-c N` resp. `-C N` operation, and test this.
4. Document in the man page that `-c N` and `-C N` override `-a` or `-u`.

This addresses issue #295. It should also help at least somewhat with issues #271 and #274.

### DIFF
--- a/src/fping.c
+++ b/src/fping.c
@@ -919,9 +919,6 @@ int main(int argc, char **argv)
         exit(1);
     }
 
-    if (alive_flag || unreachable_flag || min_reachable)
-        verbose_flag = 0;
-
     if (count_flag) {
         if (verbose_flag)
             per_recv_flag = 1;
@@ -935,6 +932,9 @@ int main(int argc, char **argv)
 
         alive_flag = unreachable_flag = verbose_flag = 0;
     }
+
+    if (alive_flag || unreachable_flag || min_reachable)
+        verbose_flag = 0;
 
     trials = (count > retry + 1) ? count : retry + 1;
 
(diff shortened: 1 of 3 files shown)


src/fping.c
```diff
@@ -2937,12 +2937,12 @@ void usage(int is_error)
     fprintf(out, "   -6, --ipv6         only ping IPv6 addresses\n");
     fprintf(out, "   -b, --size=BYTES   amount of ping data to send, in bytes (default: %d)\n", DEFAULT_PING_DATA_SIZE);
     fprintf(out, "   -B, --backoff=N    set exponential backoff factor to N (default: 1.5)\n");
-    fprintf(out, "   -c, --count=N      count mode: send N pings to each target\n");
```
schweikert, line 2940: It's a bit unfortunte that -c also changes the output format but it is definitely also changing the way probing is done. I am thinking that maybe we should either remove the option sections, or possibly introduce a new one, like "Operation mode" or something like that, where only -c, -C, and -l are documented. Otherwise, I would probably leave -c where it is now.

schweikert (commented):
Besides the minor issue of help format, this looks fine. Thanks!

src/fping.c
auerswal: I think that it would be good to have both `-c` and `-C` in the same section, because they both affect probing and output similarly.

For now, I suggest I'll redo the pull request without the change regarding where `-c` is shown in the help output.

auerswal:
I close this PR, then open a new one.

PR #297, opened 2024-02-06
More tests

auerswal:
This pull request is based on looking at the coverage report, and adding tests that should exercise more code paths. This should increase code coverage by tests.

This also contains pinging a local IPv6 multicast address. I do not know if this works with the GitHub CI. It does work on my GNU/Linux system at home, and IPv6 requires multicast support, so it should work.

If not all tests work with the GitHub CI, I'll close the pull request and look into the problems.


schweikert:
That's great, thank you!

PR #298, opened 2024-02-07
Options "-c N" / "-C N" versus "-a" and "-u" (v2)

auerswal:
1. Adjust fping -h output to emphasize that -i MSEC does not change the output format.
2. Ensure that neither -a nor -u affect -c N resp. -C N operation, and test this.
3. Document in the man page that -c N and -C N override -a or -u.

This addresses issue #295. It should also help at least somewhat with issues #271 and #274.

### DIFF
--- a/ci/test-05-options-c-e.pl
+++ b/ci/test-05-options-c-e.pl
@@ -1,6 +1,6 @@
 #!/usr/bin/perl -w
 
-use Test::Command tests => 18;
+use Test::Command tests => 33;
 
 #  -c n       count of pings to send to each target (default 1)
 #  -C n       same as -c, report results in verbose format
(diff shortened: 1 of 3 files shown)


auerswal:
I would like to adjust the `fping -h` output regarding both `-c N` and `-C N` a bit, without changing their positions:

Current:
```
   -c, --count=N      count mode: send N pings to each target
   -C, --vcount=N     same as -c, report results in verbose format
```

Suggestion:
```
   -c, --count=N      count mode: send N pings to each target and report stats
   -C, --vcount=N     same as -c, report results (not stats) in verbose format
```

Is this OK? If yes, I can add this as another commit to this pull request before you merge.

schweikert:
> I would like to adjust the `fping -h` output regarding both `-c N` and `-C N` a bit, without changing their positions:
> 
> Current:
> 
> ```
>    -c, --count=N      count mode: send N pings to each target
>    -C, --vcount=N     same as -c, report results in verbose format
> ```
> 
> Suggestion:
> 
> ```
>    -c, --count=N      count mode: send N pings to each target and report stats
>    -C, --vcount=N     same as -c, report results (not stats) in verbose format
> ```
> 
> Is this OK? If yes, I can add this as another commit to this pull request before you merge.

Sounds good to me. Thank you!

auerswal:
Thanks!

PR #301, opened 2024-02-17
Two code maintenance commits

auerswal:
* some typo fixes regarding ICMP for IPv6 sockets
* use the already existing function `stats_reset_interval()`


auerswal:
Thanks!

PR #302, opened 2024-02-21
improve help output for options -c and -C

auerswal:
As agreed in https://github.com/schweikert/fping/pull/298#issuecomment-1953720800


auerswal:
Thanks!

PR #305, opened 2024-03-03
More tests 02

auerswal:
A few more tests to both increase code coverage a little bit, and more exhaustively test `--outage`.


auerswal:
I'll rework the tests to have more lenient timing, and accept both the Linux and macOS error messages when binding an address fails, and then create a new pull request (but not right now).

PR #306, opened 2024-03-16
Initialize all ping_ts fields to -SEQMAP_TIMEOUT_IN_NS

gsnw:
Initialize all ping_ts fields to -SEQMAP_TIMEOUT_IN_NS which helps run fping immediately after a boot

PR #307, opened 2024-03-20
Check if the variable next_value->ping_ts != 0 in seqmap_add()

gsnw:
This pull request is a reduced version of @cagney and also fixes error #290 .
With this solution, however, the CPU time is faster than with pull request #306, which also fixes the error


gsnw:
The error on the test mac is the same error that occurs on azure-pipline with the macOS operating system and the current develop branch. So the cause is probably with github

schweikert:
Thank you for your work on this (exploring multiple fixes, benchmarking) !

Issue #243, opened 2022-01-31
Feature Request: Regularly Print Full Statistics Since Start


auerswal:
While I think that the current `-Q, --squiet=SECS` behavior is useful and should kept, I would like to have a variation of this theme, i.e., printing the full statistics that have accumulated so far instead of just those since the last report, optionally available, too.

## Current `-Q` Output:

The current `fping -Q1` output looks as follows:

```
$ fping -Q 1 -p 450 -c 6 127.0.0.1
[19:42:32]
127.0.0.1 : xmt/rcv/%loss = 3/3/0%, min/avg/max = 0.058/0.065/0.076
[19:42:33]
127.0.0.1 : xmt/rcv/%loss = 2/2/0%, min/avg/max = 0.058/0.058/0.059
127.0.0.1 : xmt/rcv/%loss = 6/6/0%, min/avg/max = 0.058/0.062/0.076
```

## Suggested Optional `-Q --a-new-option` Output:

I would suggest to optionally have the following different output:

```
$ fping -Q 1 --a-new-option -p 450 -c 6 127.0.0.1
[19:43:59]
127.0.0.1 : xmt/rcv/%loss = 3/3/0%, min/avg/max = 0.059/0.067/0.076
[19:44:00]
127.0.0.1 : xmt/rcv/%loss = 5/5/0%, min/avg/max = 0.058/0.063/0.076
127.0.0.1 : xmt/rcv/%loss = 6/6/0%, min/avg/max = 0.058/0.062/0.076
```

If used together with the `-o, --outage` option, this would regularly show the *accumulated* outage times of hosts during, e.g., network changes.

I have created a proof-of-concept implementation that introduces the new option `-F, --full` to provide this feature. Adding `-F` to a command line with `-Q` does not regularly print *interval* statistics, but instead regularly prints *full* statistics (since the start of `fping`). I'll add a pull request for this proof-of-concept implementation in a moment and add it in a new comment to this issue.

I am open to using different option names, of course.

What do you think?

Best regards,
Erik

auerswal:
I have closed pull request #244 since there has not been any feedback so far. Perhaps you would like to close the issue as *won't fix*?

schweikert:
I like the idea and I would still like the feature to be integrated. I will provide more timely feedback if you re-open or recreate a pull request.

Instead of adding another option, what do you think about extending the syntax of the --squiet option? It could be: `--squiet=60,cumulative`, for example.

PR #315, opened 2024-04-30
Print cumulative stats with -Q SECS,cumulative

auerswal:
With -Q SECS, fping emits information for the just finished interval.  This is good for long-running fping processes where one would like to get an update on recent results, ignoring missing responses from older intervals.  This seems appropriate for use with other software like netdata or smokeping.

But for short(er)-running fping processes, e.g., during a change window, it may be more interesting to get a status update since the beginning of the change window, i.e., the start of fping. Adding a comma followed by the keyword "cumulative" to -Q SECS (i.e., -Q SECS,cumulative) changes the interim reports to output the current per system overall statistics, unless -N is used, too.

The new syntax is documented in fping.pod and added to fping -h output.

Four tests are added:

  1. Characters after SECS are ignored as before.
  2. Unknown keywords are ignored as before.
  2. Adding ",cumulative" changes -Q output.
  3. Adding ",cumulative" also affects the additional -o output.

This addresses issue #243.

### DIFF
--- a/ci/test-08-options-n-q.pl
+++ b/ci/test-08-options-n-q.pl
@@ -1,6 +1,6 @@
 #!/usr/bin/perl -w
 
-use Test::Command tests => 24;
+use Test::Command tests => 36;
 
 #  -n         show targets by name (-d is equivalent)
 #  -O n       set the type of service (tos) flag on the ICMP packets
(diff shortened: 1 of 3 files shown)


auerswal:
Thanks!

PR #318, opened 2024-05-18
Debug with Visual Studio Code

ghost:
The basis for debugging with Visual Studio Code has been created. This may not yet be perfect, but it may be a start.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,35 @@
+{
+  "configurations": [
+    {
+      "name": "Debug fping",
+      "type": "cppdbg",
+      "request": "launch",
+      "program": "${workspaceFolder}/src/fping",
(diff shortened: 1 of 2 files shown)


.vscode/launch.json
```diff
@@ -0,0 +1,35 @@
+{
+  "configurations": [
+    {
+      "name": "Debug fping",
+      "type": "cppdbg",
+      "request": "launch",
+      "program": "${workspaceFolder}/src/fping",
+      "args": ["8.8.8.8"],
```
auerswal: Perhaps use 127.0.0.1 instead?

gsnw: I have changed it

auerswal:
I do not use Visual Studio Code, so I cannot assess if this works as intended, but I do not notice anything that looks problematic.

Issue #310, opened 2024-03-28
-D, --timestamp with custom time string eq: %Y-%m-%d %H:%M:%S


coolshou:
Currently the --timestamp output string as following, can we custom it in human readable format  eq: %Y-%m-%d %H:%M:%S
fping -D -l 192.168.0.21
[1711612145.46829] 192.168.0.21 : [0], 64 bytes, 27.2 ms (27.2 avg, 0% loss)
[1711612146.48808] 192.168.0.21 : [1], 64 bytes, 46.0 ms (36.6 avg, 0% loss)
[1711612147.46816] 192.168.0.21 : [2], 64 bytes, 26.1 ms (33.1 avg, 0% loss)

schweikert:
We can't change the format, so this would mean adding yet another option (and in general we have too many already). We could possibly add --timestamp-format (without a short-option equivalent), if somebody wants to create a pull request for that.

gsnw:
If I see this correctly, the one function in `optparse.c` would first have to be created or extended so that the long name of the comman line option can be used without short.
Currently the short is always returned as ASCII int.

auerswal:
I did a quick test, a number greater than 255 can be used instead of a character for options that only have a long name:

```
diff --git a/src/fping.c b/src/fping.c
index c58e20c..f7a3129 100644
--- a/src/fping.c
+++ b/src/fping.c
@@ -552,6 +552,7 @@ int main(int argc, char **argv)
         { "version", 'v', OPTPARSE_NONE },
         { "reachable", 'x', OPTPARSE_REQUIRED },
         { "fast-reachable", 'X', OPTPARSE_REQUIRED },
+        { "long-only-test", 256, OPTPARSE_NONE },
 #if defined(DEBUG) || defined(_DEBUG)
         { NULL, 'z', OPTPARSE_REQUIRED },
 #endif
@@ -872,6 +873,11 @@ int main(int argc, char **argv)
             outage_flag = 1;
             break;
 
+        case 256:
+            puts("long-only option given");
+            exit(0);
+            break;
+
         case '?':
             fprintf(stderr, "%s: %s\n", argv[0], optparse_state.errmsg);
             fprintf(stderr, "see 'fping -h' for usage information\n");
```

This works as intended for the long option:

```
$ ./src/fping --long-only-test
long-only option given
```

`#define`s or an `enum` could be used instead of simple numbers.

[...]

PR #321, opened 2024-06-01
Predefined various timestamp formats

ghost:
This extension allows you to change the format of the timestamp when using `fping -D -l 127.0.0.1`.
Example: `fping -D -l --timestamp-format=ctime 127.0.0.1`

Issue: #310

### DIFF
--- a/ci/test-05-options-c-e.pl
+++ b/ci/test-05-options-c-e.pl
@@ -1,6 +1,6 @@
 #!/usr/bin/perl -w
 
-use Test::Command tests => 33;
+use Test::Command tests => 48;
 
 #  -c n       count of pings to send to each target (default 1)
 #  -C n       same as -c, report results in verbose format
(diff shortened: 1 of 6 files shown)


src/fping.c
```diff
@@ -2992,6 +3036,7 @@ void usage(int is_error)
     fprintf(out, "   -C, --vcount=N     same as -c, report results (not stats) in verbose format\n");
     fprintf(out, "   -d, --rdns         show targets by name (force reverse-DNS lookup)\n");
     fprintf(out, "   -D, --timestamp    print timestamp before each output line\n");
+    fprintf(out, "       --timestamp-format=FORMAT  show timestamp using the given format (-D are required): [ctime|iso|reltime]\n");
```
schweikert: `-D is required` or `-D required` to shorten it.

schweikert: To shorten it a bit more:

show timestamp in the given format (-D required): ctime|iso|reltime\n

So maybe wrapping can be avoided.

schweikert (approved):

src/fping.c
gsnw: Thanks for the feedback and the suggestion.
I have adopted the change and hope the rest of the code is also ok

ci/test-05-options-c-e.pl
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/perl -w
 
-use Test::Command tests => 33;
+use Test::Command tests => 42;
```
auerswal: Please consider adding two error test cases for the new option:

1. omitting the format name, and
2. using an unsupported format name.

src/fping.c
```diff
@@ -2939,6 +2955,34 @@ void ev_remove(struct event_queue *queue, struct event *event)
...
+{
+    char time_buffer[100];
+    time_t current_time_s;
+    struct tm *local_time;
+
+    if(timestamp_format >= 1 && timestamp_format <= 3) {
+        current_time_s = current_time_ns / 1000000000;
+        local_time = localtime(&current_time_s);
+        if(timestamp_format == 1) {
+          strftime(time_buffer, sizeof(time_buffer), "%c", local_time);
```
auerswal: Before this patch, `strftime` is not used in the `fping` code. Should `configure` check for availability of `strftime`?

src/fping.c
```diff
@@ -2939,6 +2955,34 @@ void ev_remove(struct event_queue *queue, struct event *event)
...
+void print_timestamp_format(int64_t current_time_ns, int timestamp_format)
+{
+    char time_buffer[100];
+    time_t current_time_s;
+    struct tm *local_time;
+
+    if(timestamp_format >= 1 && timestamp_format <= 3) {
+        current_time_s = current_time_ns / 1000000000;
+        local_time = localtime(&current_time_s);
+        if(timestamp_format == 1) {
```
auerswal: I'd think this part (i.e., the inner `if`) would be more readable using `switch`.

auerswal (approved):
All in all, this looks OK. Please consider my comments, but you do not need to make any changes on my behalf.

Thanks!

auerswal:
Please also consider adding a description to the man page, i.e., `fping.pod`. Thanks!

auerswal:
I am not so sure any more, especially regarding `reltime`. With `dmesg`, `--time-format=reltime` provides *relative* time stamps for short time differences, and absolute time stamps for longer gaps, e.g.:

```
[Jun 9 21:17] …
[ +21,801486] …
[  +0,000004] …
[  +0,000004] …
[  +0,000002] …
[  +0,000001] …
[  +0,091934] …
[  +0,000003] …
[  +0,000002] …
[  +0,000002] …
[  +0,000001] …
[  +0,000004] …
[  +0,574391] …
[Jun 9 22:54] …
```

Perhaps we should leave out `reltime` for now?

And a nit: both `date --iso-8601` and `dmesg --time-format=iso` print the timezone offset with a colon between hours and minutes (`HH:MM`). The `%z` format specifier does not include the colon, but I did not see a better format specifier in `man strftime`.

(As another example, GNU date has two quite similar "named" time format options where the output differs mainly with the character between date and time, either a space or a `T`: `date --rfc-3339=s` and `date --iso-8601=s` [`--rfc-3339` does not use the `locale` settings, `--iso-8601` does, but `fping` does not use `locale` settings (IMHO `fping` is better in that regard)].)

ci/test-05-options-c-e.pl
gsnw: I have add two negativ test `--timestamp-format` and `--timestamp-format="%Y-%m-%d %H:%M:%S"` both ending in function usage() with exit code 1

src/fping.c
gsnw: The if statement has been completely converted into a switch statement. Many thanks for this feedback.
Switch statement with the default is also better than the old if else.
This way, the default is always used if the function is passed incorrectly

gsnw:
The formats are not perfect. I have now turned reltime into rfc3339, even with the knowledge that the format may not be one hundred percent perfect.
However, it corresponds to the example from the ticket.

[1] https://medium.easyread.co/understanding-about-rfc-3339-for-datetime-formatting-in-software-engineering-940aa5d5f68a?gi=fc35d513c0fd

src/fping.c
gsnw: It's also done

auerswal (approved):
Thanks for the adjustments, this looks good. :-)

auerswal:
We haven't talked about the process for merging pull requests yet. I would be willing to merge this pull request next weekend, unless there are objections. Is that OK?

gsnw:
The last github merges were probably done with rebase.
Everything has advantages and disadvantages.

All my pull requests are always based on the current state of develop and I always try to do a rebase on all branches in my repository when the develop branch changes here.

So I would prefer rebase and merge

auerswal:
Ah, I see, that is one of the options shown when clicking the arrow next to *Merge pull request*. I think I (slightly) prefer that option too, because automatic merge commits add some noise, but only little information. I'll go ahead and do that, thanks!

(For the record: I do *not* like the *Squash and merge* option, I prefer to keep the individual commits.)

PR #323, opened 2024-06-14
Fix incorrect return-value check for a scanf like function (CWE-253)

ghost:
This pull request fix the codeql security CWE-253 alert `Incorrect return-value check for a 'scanf'-like function`

### DIFF
--- a/src/fping.c
+++ b/src/fping.c
@@ -621,7 +621,7 @@ int main(int argc, char **argv)
             break;
 
         case 't':
-            if (!sscanf(optparse_state.optarg, "%f", &opt_value_float))
+            if (sscanf(optparse_state.optarg, "%f", &opt_value_float) != 1)
                 usage(1);
             if (opt_value_float < 0) {
                 usage(1);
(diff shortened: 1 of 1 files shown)


auerswal (approved):
I have looked at the changes, and reviewed the `sscanf` man page. This change seems correct, because only the return value of `1` represents matching the single expected value. 

A value of 0 would mean that nothing matched the format string. The current code checks against this. A number greater than 1 seems unlikely with a format string that comprises a single conversion specification. In case of an error, `EOF` could also be returned, but the current code would not catch this, I think.

gsnw:
Fix issue #324

auerswal:
`sscanf()` returns `EOF` when used with an empty string as input. This can be given as option argument, e.g., using `fping -t '' …`. Without the pull request, `fping` does not catch this invalid option argument.

auerswal:
I have looked at the changes again, IMHO they are "obviously correct", so I'll just go ahead and merge this.
@schweikert: Please let me know if this is OK.

schweikert:
Looks good to me as well. Thanks!

PR #327, opened 2024-06-23
Some additional tests

auerswal:
These new tests check more aspects of `fping` behavior and make two existing tests a little bit more reliable, but they are not intended to increase code coverage.


auerswal:
Hm, the test failing now is one of the tests I tried to make more reliable. Perhaps Those tests should just be skipped on macOS.

PR #328, opened 2024-06-23
ci: skip an unreliable test on macOS

auerswal:
Sometimes, this test fails on macOS, at least on the GitHub CI. It looks as if the fping process is killed before it produces the required output.  This test works reliably on GNU/Linux, so keep it, but skip it when running on macOS.

### DIFF
--- a/ci/test-07-options-i-m.pl
+++ b/ci/test-07-options-i-m.pl
@@ -51,7 +51,10 @@ $cmd->stderr_like(qr{\[\d+:\d+:\d+\]
 }
 
 # fping -l -Q
-{
+SKIP: {
+if($^O eq 'darwin') {
+    skip 'On macOS, this test is unreliable', 2;
+}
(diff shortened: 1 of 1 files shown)


schweikert (approved):

gsnw:
I use Azure Devops with macOS and have no problem with `ci/test-07-options-i-m.pl`.
We should investigate why this happens with Github Action via a ticket if necessary

ghost (approved):

auerswal:
I have no experience with macOS except for the fping GitHub CI. The macOS tests on this GitHub CI have proven problematic in the past, where I had to adjust the timing of tests to make them work. Since I have seen some failures for this test, but have also seen it succeed (re-running the failed tests usually makes it pass), I would prefer not to have to look at those spurious test failures.

I would expect the problem to be caused by high load on the test runners.

I do not have the motivation to find out if there is any support regarding the CI infrastructure and, if it exists, if it is useful.

@gsnw: Do you know a way to skip this test only on the GitHub CI, but not on Azure DevOps?

gsnw:
You can merge this pull-request. I have create a new issue.

PR #331, opened 2024-07-07
Generator limit

### LINKED ISSUE #299: Inconsistent limits for address generation via `-g`, `--generator` using either range or CIDR
When using `fping -g` with a range, the number of addresses that may be generated is limited (via the `MAX_GENERATE` constant in `src/fping.c`):

```
$ fping -q --stats -g 127.0.0.1 127.1.255.254
./src/fping: -g parameter generates too many addresses
```

But this limit is not applied when using CIDR notation:

```
$ ipcalc 127.0.0.0/15 | grep 'HostM[ai][nx]'
HostMin:      127.0.0.1
HostMax:      127.1.255.254
$ fping -q --stats -g 127.0.0.0/15

  131070 targets
  131070 alive
       0 unreachable
       0 unknown addresses

       0 timeouts (waiting for response)
  131070 ICMP Echos sent
  131070 ICMP Echo Replies received
       0 other ICMP received

 0.005 ms (min round trip time)
 0.033 ms (avg round trip time)
 0.166 ms (max round trip time)
     1328.103 sec (elapsed real time)
```

I have mentioned this in a [comment][1] on issue #59, because I think that applying a consistent limit to address generation is a prerequisite for expanding `-g` functionality to include IPv6, but I'd prefer to track this issue independently.

[1]: https://github.com/schweikert/fping/issues/59#issuecomment-1078896614

I can refresh my respective patches that have been part of the IPv6 generator pull request #254 and submit them as a new pull request. Just let me know if this desired.

auerswal:
This pull request addresses GitHub issue #299 by:

* using the same maximum number of generated addresses for both range and CIDR arguments,
* fixing the off-by-one error in the generator limit, and
* increasing the limit to allow a 15-bit IPv4 prefix as generator input, or a range of the same size.

The generator limit is not exactly tested via CI, because pinging 100000 (or 100001) localhost IPv4 addresses takes more than 15 minutes on my computer. Testing 2^17-2 localhost IPv4 addresses takes more than 20 minutes on my computer. Thus I have tested this only manually.

### DIFF
--- a/ci/test-06-options-f-h.pl
+++ b/ci/test-06-options-f-h.pl
@@ -1,6 +1,6 @@
 #!/usr/bin/perl -w
 
-use Test::Command tests => 72;
+use Test::Command tests => 75;
 use Test::More;
 use File::Temp;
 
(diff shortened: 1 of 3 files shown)


schweikert (approved):
Looks good to me, thank you!

PR #332, opened 2024-07-12
new option --check-source

auerswal:
Using the new option --check-source discards Echo reply packets sourced from an address that is different from the specified target address.  This can be useful on busy monitoring hosts, because it is possible that two fping processes running at he same time use the same ICMP Echo Reply Identifier for ICMP Echo messages sent to different hosts.

This aims to address GH issue #206.

### DIFF
--- a/ci/test-02-help.pl
+++ b/ci/test-02-help.pl
@@ -1,6 +1,6 @@
 #!/usr/bin/perl -w
 
-use Test::Command tests => 12;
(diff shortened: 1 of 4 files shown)


ghost (approved):
I see no problem in adopting this

auerswal:
I do not understand why the test fails on the GitHub CI, it works on my Ubuntu 20.04 system. I am closing this pull request.

PR #334, opened 2024-07-13
new option --check-source

auerswal:
Using the new option --check-source discards Echo reply packets sourced from an address that is different from the specified target address.  This can be useful on busy monitoring hosts, because it is possible that two fping processes running at he same time use the same ICMP Echo Reply Identifier for ICMP Echo messages sent to different hosts.

This aims to address GH issue #206.


ghost (approved):

auerswal:
I plan to merge this next weekend, unless there is any objection.

Issue #68, opened 2014-09-11
Future request : print returned TOS value


GBert:
Ping with TOS field is already in the code, but printing the result is missing. I know that you don't like to change the output so I have added tos_flag which is set by the -O flag
(https://github.com/GBert/fping/commit/32e266efec137f19437a845c3ed3ce4d77a035c5)

```
--- a/src/fping.c
+++ b/src/fping.c
@@ -230,6 +230,7 @@ typedef struct host_entry
      int                  min_reply_i;        /* shortest response time */
      int                  total_time_i;       /* sum of response times */
      int                  *resp_times;        /* individual response times */
+     int                  tos;                /* individual tos field */
 #if defined( DEBUG ) || defined( _DEBUG )
      int                  *sent_times;        /* per-sent-ping timestamp */
 #endif /* DEBUG || _DEBUG */
@@ -295,6 +296,7 @@ struct timezone tz;

 /* switches */
 int generate_flag = 0;              /* flag for IP list generation */
+int tos_flag = 0;                        /* flag if tos should be printed */
 int verbose_flag, quiet_flag, stats_flag, unreachable_flag, alive_flag;
 int elapsed_flag, version_flag, count_flag, loop_flag;
 int per_recv_flag, report_all_rtts_flag, name_flag, addr_flag, backoff_flag;
@@ -363,7 +365,7 @@ int main( int argc, char **argv )
     int c, i, n;
     char *buf;
     uid_t uid;
-    int tos = 0;
+    int tos = 0;
     HOST_ENTRY *cursor;

     s = open_ping_socket();
@@ -548,6 +550,7 @@ int main( int argc, char **argv )
                 if ( setsockopt(s, IPPROTO_IP, IP_TOS, &tos, sizeof(tos))) {
                     perror("setting type of service octet IP_TOS");
                 }
+               tos_flag = 1;
             }
             else {
                 usage(1);
@@ -1605,6 +1608,7 @@ int wait_for_reply(long wait_time)
         sum_replies += this_reply;
         h->total_time += this_reply;
         h->total_time_i += this_reply;
+        h->tos = ip->ip_tos;
         total_replies++;
     }

@@ -1664,7 +1668,10 @@ int wait_for_reply(long wait_time)
             printf( "%s", h->host );

             if( verbose_flag )
-                printf( " is alive" );
+                if ( tos_flag )
+                    printf( " is alive (TOS %d)", h->tos );
+                else
+                    printf( " is alive" );

             if( elapsed_flag )
[...]

PR #335, opened 2024-07-14
Print returned TOS value

ghost:
Added tos to the header again, but only for IPv4.
IPv6 uses ds field here
Currently, TOS is only output in one place as in the ticket example.

Issue #68

### DIFF
--- a/doc/fping.pod
+++ b/doc/fping.pod
@@ -186,6 +186,11 @@ Calculate "outage time" based on the number of lost pings and the interval used 
 Set the typ of service flag (TOS). I<N> can be either decimal or hexadecimal
 (0xh) format.
 
+=item B<--print-tos>
+
+Displays the TOS value in the output, if TOS is not present, "(TOS unknown)" is returned.
+IPv4 only, requires root privileges or cap_net_raw.
+
 =item B<-p>, B<--period>=I<MSEC>
 
 In looping or counting modes (B<-l>, B<-c>, or B<-C>), this parameter sets
--- a/src/fping.c
+++ b/src/fping.c
@@ -361,6 +361,7 @@ int timestamp_format_flag = 0;
 int random_data_flag = 0;
 int cumulative_stats_flag = 0;
 int check_source_flag = 0;
+int print_tos_flag = 0;
 #if defined(DEBUG) || defined(_DEBUG)
 int randomly_lose_flag, trace_flag, print_per_system_flag;
 int lose_factor;
@@ -558,6 +559,7 @@ int main(int argc, char **argv)
         { "reachable", 'x', OPTPARSE_REQUIRED },
         { "fast-reachable", 'X', OPTPARSE_REQUIRED },
         { "check-source", '0', OPTPARSE_NONE },
+        { "print-tos", '0', OPTPARSE_NONE },
 #if defined(DEBUG) || defined(_DEBUG)
         { NULL, 'z', OPTPARSE_REQUIRED },
 #endif
@@ -569,17 +571,19 @@ int main(int argc, char **argv)
         switch (c) {
         case '0':
             if(strstr(optparse_state.optlongname, "timestamp-format") != NULL) {
-              if(strcmp(optparse_state.optarg, "ctime") == 0) {
-                timestamp_format_flag = 1;
-              }else if(strcmp(optparse_state.optarg, "iso") == 0) {
-                timestamp_format_flag = 2;
-              }else if(strcmp(optparse_state.optarg, "rfc3339") == 0) {
-                timestamp_format_flag = 3;
-              }else{
-                usage(1);
-              }
+                if(strcmp(optparse_state.optarg, "ctime") == 0) {
+                  timestamp_format_flag = 1;
+                }else if(strcmp(optparse_state.optarg, "iso") == 0) {
+                  timestamp_format_flag = 2;
+                }else if(strcmp(optparse_state.optarg, "rfc3339") == 0) {
+                  timestamp_format_flag = 3;
+                }else{
+                  usage(1);
+                }
             } else if (strstr(optparse_state.optlongname, "check-source") != NULL) {
                 check_source_flag = 1;
+            } else if (strstr(optparse_state.optlongname, "print-tos") != NULL) {
+                print_tos_flag = 1;
             } else {
                 usage(1);
             }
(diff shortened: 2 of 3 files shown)


gsnw:
The macOS test works perfectly under AzurePipline. It is therefore a Github problem

src/fping.c
```diff
@@ -2153,14 +2161,16 @@ int decode_icmp_ipv4(
...
+    unsigned short *seq,
+    int *ip_header_tos)
 {
     struct icmp *icp;
+    struct ip *ip;
     int hlen = 0;
 
-    if (!using_sock_dgram4) {
-        struct ip *ip = (struct ip *)reply_buf;
+    ip = (struct ip *)reply_buf;
```
auerswal: I do not think this is correct. It seems to me that only a raw socket provides the IP header as part of the received data. The code after the `if (…) {…}` assumes that the ICMP datagram starts directly at the beginning of `recv_buf` when using a datagram socket.

gsnw: In my opinion, the code is correct.
Both sockets return an ip header with the TOS value.

* Type Of Service == TOS

**SOCK_RAW**
```
IP Header:
  |-Version        : 4
  |-Internet Header Length  : 5 DWORDS or 20 Bytes
  |-Type Of Service   : 0
  |-Total Length   : 84  Bytes
  |-Identification  : 58676
  |-Time To Live    : 64
  |-Protocol     : 1
  |-Header Checksum : 22386
  |-Source IP       : 127.0.0.1
  |-Destination IP  : 127.0.0.1
```

**SOCK_DGRAM**
```
IP Header:
  |-Version        : 7
  |-Internet Header Length  : 4 DWORDS or 16 Bytes
  |-Type Of Service   : 101
  |-Total Length   : 29556  Bytes
  |-Identification  : 2560
  |-Time To Live    : 0
  |-Protocol     : 0
  |-Header Checksum : 0
  |-Source IP       : 0.0.0.0
  |-Destination IP  : 0.0.0.0
```

I checked this in advance in a separate program

auerswal: This shows that the data received from a `SOCK_DGRAM` socket does *not* start with an IP header:
 * The valid versions for IP headers are 4 or 6, not 7. Since we are handling ICMP via IPv4 the version must be 4.
 * The minimum IPv4 header length is 5 (20 bytes), not 4 (16 bytes)
 * The protocol is given as 0, but the protocol number for ICMP for IPv4 is 1

The code finds the ICMP data at offset `hlen` from the start of the data `reply_buf`:
```
icp = (struct icmp *)(reply_buf + hlen);
```
In case of `using_sock_dgram4`, `hlen` equals 0, and the ICMP data starts at the beginning of `reply_buf`. In case of `!using_sock_dgram4`, the value of `hlen` is extracted from the IP header at the start of `reply_buf` in order to skip it to get at the ICMP data.

The data can either start with an IPv4 header followed by ICMP, or start with ICMP, but not both at the same time.

auerswal:
I like the general approach of using an option to enable printing of received TOS values. I also like to approach of starting with the default output (`<target> is alive (TOS <value>)`).

src/fping.c
auerswal: You could do the following:
* in `decode_icmp_ipv4()`:
  * initialize the TOS value to -1 (`*ip_header_tos = -1;`)
  * set it to the received value inside the `if (!using_sock_dgram4) {…}` block
* in `wait_for_reply()`:
  * only print the received TOS value if it is greater than -1 (you could print `(TOS unknown)` or just not add the output)

gsnw:
I have changed the logic a bit and --print-tos can be used independently of the -O parameter.
However, it is only displayed if the value is present in the IP header.

doc/fping.pod
```diff
@@ -179,6 +179,10 @@ Calculate "outage time" based on the number of lost pings and the interval used
 Set the typ of service flag (TOS). I<N> can be either decimal or hexadecimal
 (0xh) format.
 
+=item B<--print-tos>
+
+Displays the TOS value in the output. However, the value is only displayed if a TOS value is present in the IP header.
```
auerswal: Any IPv4 packet always has a TOS byte value (or rather DSCP + ECN field values nowadays) in the IP header. But `fping` can only read when using raw sockets, and currently only for IPv4. I would suggest to write something like: "IPv4 only, requires root privileges or cap_net_raw."

src/fping.c
```diff
@@ -2153,13 +2159,15 @@ int decode_icmp_ipv4(
...
-    unsigned short *seq)
+    unsigned short *seq,
+    int *ip_header_tos)
 {
     struct icmp *icp;
     int hlen = 0;
 
     if (!using_sock_dgram4) {
         struct ip *ip = (struct ip *)reply_buf;
+        *ip_header_tos = ip->ip_tos;
```
auerswal, line 2174: This looks good, thanks!

src/fping.c
```diff
@@ -2486,8 +2496,11 @@ int wait_for_reply(int64_t wait_time)
         if (verbose_flag || alive_flag) {
             printf("%s", h->host);
 
-            if (verbose_flag)
+            if (verbose_flag) {
                 printf(" is alive");
+                if(print_tos_flag && ip_header_tos != -1)
+                    printf(" (TOS %d)", ip_header_tos);
+            }
```
auerswal, line 2519: I think this is OK. Alternatively, with `print_tos_flag && ip_header_tos == -1`, you could print ` (TOS unknown)`. I am not sure what is better, your variant is OK as it is.

src/fping.c
```diff
@@ -3057,5 +3070,6 @@ void usage(int is_error)
     fprintf(out, "   -v, --version      show version\n");
     fprintf(out, "   -x, --reachable=N  shows if >=N hosts are reachable or not\n");
     fprintf(out, "   -X, --fast-reachable=N exits true immediately when N hosts are found\n");
+    fprintf(out, "       --print-tos    show tos value (only if present in the IP header)");
```
auerswal: I would rather omit the part in parentheses.

auerswal (changes requested):
Please adjust the documentation (man page and help output), and adjust the code to the changes by my just merged pull request #336.

src/fping.c
gsnw: I have changed it

auerswal (approved):
Thanks! Looks good to me!

gsnw-sebast:
Perfect, can you merge that please? I made a few changes to my github account and lost my permissions as a result

PR #336, opened 2024-07-15
Sets SEQMAP_TIMEOUT_IN_NSSEQMAP_TIMEOUT_IN_NS as INT64_C

ghost:
Fix macOS build warning

```
seqmap.c:86:13: warning: format specifies type 'long long' but the argument has type 'long' [-Wformat]
            SEQMAP_TIMEOUT_IN_NS, host_nr, ping_count, seqmap_next_id);
            ^~~~~~~~~~~~~~~~~~~~
seqmap.c:60:30: note: expanded from macro 'SEQMAP_TIMEOUT_IN_NS'
#define SEQMAP_TIMEOUT_IN_NS 10000000000
                             ^~~~~~~~~~~
1 warning generated.
```

### DIFF
--- a/src/seqmap.c
+++ b/src/seqmap.c
@@ -57,7 +57,7 @@
 static SEQMAP_VALUE* seqmap_map = NULL;
 static unsigned int seqmap_next_id = 0;
 
-#define SEQMAP_TIMEOUT_IN_NS 10000000000
+#define SEQMAP_TIMEOUT_IN_NS INT64_C(10000000000)
 #define SEQMAP_UNASSIGNED_HOST_NR UINT_MAX
 
 void seqmap_init()


auerswal:
Thanks for noticing this and looking into it, @gsnw!

How about using the `INT64_C(…)` macro to specify the type of the constant instead of using a cast to `int64_t`?

On the affected macOS system, the type suffix for the numeric constant would need to be `LL` to avoid the warning, but on an x86_64 GNU/Linux system the type suffix should be `L`. The `INT64_C(…)` macro is intended to choose the correct suffix for the given system.

The following patch works on my x86_64 GNU/Linux system:

```
diff --git a/src/seqmap.c b/src/seqmap.c
index 62506d0..7675bc1 100644
--- a/src/seqmap.c
+++ b/src/seqmap.c
@@ -57,7 +57,7 @@
 static SEQMAP_VALUE* seqmap_map = NULL;
 static unsigned int seqmap_next_id = 0;
 
-#define SEQMAP_TIMEOUT_IN_NS 10000000000
+#define SEQMAP_TIMEOUT_IN_NS INT64_C(10000000000)
 #define SEQMAP_UNASSIGNED_HOST_NR UINT_MAX
 
 void seqmap_init()
```

gsnw:
Ah thanks, I've learned something again.
I didn't know that yet. I've changed it and tested it.

auerswal (approved):

PR #337, opened 2024-07-23
Document option incompatiblities

moritz:
The documentation for `--count` already mentions that it modifies the output.

This changes the documentation for `--alive` and `--unreach` to explicitly state that they don't work together with `--count`.

### DIFF
--- a/doc/fping.pod
+++ b/doc/fping.pod
@@ -38,6 +38,8 @@ Restrict name resolution and IPs to IPv6 addresses.
 
 Show systems that are alive.
 
+Not compatible with option B<--count> / B<-c>.
+
 =item B<-A>, B<--addr>
 
 Display targets by address rather than DNS name. Combined with -d, the output
@@ -237,6 +239,8 @@ Ignored (for compatibility with fping 2.4).
 =item B<-u>, B<--unreach>
 
 Show targets that are unreachable.
+
+Not compatible with option B<--count> / B<-c>.
 
 =item B<-v>, B<--version>
 


doc/fping.pod
```diff
@@ -38,6 +38,8 @@ Restrict name resolution and IPs to IPv6 addresses.
 
 Show systems that are alive.
 
+Not compatible with option B<--count> / B<-c>.
```
auerswal, line 41: How about adding something like "(Option B<-c> or B<-C> overrides B<-a>)"? This could start in the same line as the existing description.

doc/fping.pod
```diff
@@ -238,6 +240,8 @@ Ignored (for compatibility with fping 2.4).
 
 Show targets that are unreachable.
 
+Not compatible with option B<--count> / B<-c>.
```
auerswal, line 243: How about adding something like "(Option B<-c> or B<-C> overrides B<-u>)"? This could start in the same line as the existing description.

auerswal (changes requested):
Thanks for the pull request!

I'd say that the documentation changes should account for more details:
 * Both options -c and -C override -a and -u.
 * Using "incompatible" is rather vague, "override" is more precise.
 * References to other options in the fping man page usually use the short version only (if applicable).
 * I would rather have this added to the same line as the existing description instead of putting it in a new paragraph.

Would you like to adjust your pull request, or would you prefer me to implement my suggestion?

moritz:
I'd be happy if you adjusted it to your preference before merging.

auerswal:
This has been replaced by merging #338.

PR #340, opened 2024-07-28
Another batch of additional tests

auerswal:
* fix test of unprivileged use
* test `--print-tos` output with IPv4 datagram socket
* test `--print-tos` output with IPv6
* test `--print-tos` output without `-O`
* test `-c` and `-C` with reply source address different from target address (using `ff02::1`)
* test `-d` a little, since it may be mostly equivalent to `-n`, but still uses different code paths
* test error message and statistics output when name resolution does not provide an IP address

This has been tested on GNU/Linux. I do not know if it also works on macOS.

### DIFF
--- a/ci/test-01-basics.pl
+++ b/ci/test-01-basics.pl
@@ -1,6 +1,6 @@
 #!/usr/bin/perl -w
 
-use Test::Command tests => 12;
+use Test::Command tests => 15;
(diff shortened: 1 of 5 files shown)


auerswal:
Well, I forgot to skip (at least some) IPv6 tests, resulting in macOS test failures.

PR #342, opened 2024-08-04
fix -k, --fwmark with setuid fping executable

auerswal:
* Setting the SO_MARK socket option requires root privileges (or CAP_NET_ADMIN), thus temporarily elevate privileges for this operation, if possible, by using the p_setsockopt() function. This allows to use -k, --fwmark with setuid fping.
* Adjust README.md since setuid now works for -k, --fwmark.
* Remove "sudo" from testing -k since it is no longer needed.
* Test failure of setting firewall mark without privileges or capabilities for both IPv4 and IPv6.


gsnw-sebast:
That looks good

auerswal:
Thanks for the review, I plan to merge this next weekend.

PR #343, opened 2024-08-11
CI: use File::Temp to create temporary directory

### LINKED ISSUE #339: Test `ci/test-11-unpriv.pl` uses hard-coded copy destination and does not clean up afterwards
The test `ci/test-11-unpriv.pl` uses a hard-coded destination of `/tmp/fping.copy` to create a binary with neither setuid bit nor capabilities. This copy is not removed, it remains after the tests have finished.

It would be preferable to use a randomized name, or to put the copy inside a temporary directory with a randomized name, using, e.g., mktemp(1) or equivalent Perl functionality. This file or directory should be deleted after testing to avoid accumulating more and more such copies.

auerswal:
Before, ci/test-11-unpriv.pl used a hard-coded file name of /tmp/fping.copy to create an unprivileged copy of fping.  This file was not deleted after it was no longer used.

Now, the unprivileged fping copy is created in a temporary directory that is deleted after all tests from the file.

This addresses GH issue #339.

### DIFF
--- a/ci/test-11-unpriv.pl
+++ b/ci/test-11-unpriv.pl
@@ -1,6 +1,7 @@
 #!/usr/bin/perl -w
 
 use English;
+use File::Temp qw/ tempdir /;
 use Test::Command;
 use Test::More;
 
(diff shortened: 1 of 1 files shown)


gsnw-sebast:
Considering system independence, I would suggest using `File::Copy` instead of `system()`

```
use File::Copy;

copy("datafile.dat", "datafile.bak") or die "copy failed: $!";
```

The chmod function should also be available directly in perl (https://perldoc.perl.org/functions/chmod)

gsnw-sebast:
You can merge, everything looks good

auerswal:
Thanks for the review. I'll go ahead and merge this now.

PR #345, opened 2024-08-21
Print returned TTL value

nalves599:
Print the returned TTL value from the IP header.

### DIFF
--- a/ci/test-04-options-a-b.pl
+++ b/ci/test-04-options-a-b.pl
@@ -1,6 +1,6 @@
 #!/usr/bin/perl -w
 
-use Test::Command tests => 32;
+use Test::Command tests => 41;
 use Test::More;
 use Time::HiRes qw(gettimeofday tv_interval);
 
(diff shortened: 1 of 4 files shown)


auerswal:
Thanks for this pull request! I have started looking at the changes and will provide a first review soon.

ci/test-04-options-a-b.pl
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/perl -w
 
-use Test::Command tests => 32;
+use Test::Command tests => 35;
```
auerswal: The tests are counted differently: you have added 9 tests (each of the added `exit_is_num()`, `stdout_like()`, `stderr_is_eq()` is counted as one test). Please adjust this.

ci/test-11-unpriv.pl
```diff
@@ -40,7 +40,7 @@ sub get_ping_gid_range {
 }
 
 sub test_unprivileged_works {
-    plan tests => 12;
+    plan tests => 13;
```
auerswal: You have added 3 tests, not one (see above for the explanation).

doc/fping.pod
```diff
@@ -140,6 +140,11 @@ Print usage message.
 
 Set the IP TTL field (time to live hops).
 
+=item B<--print-ttl>
+
+Displays the TTL value in the output, if TTL is not present, "(TTL unknown)" is returned.
```
auerswal: Every IP packet has a TTL (called *Hop Count* for IPv6, but the IPv4 TTL field is de facto a hop count anyway). The issue is that `fping` can only read it when using raw sockets, and only for IPv4. I would prefer the man page to description to reflect this. I have just created pull request #346 to adjust this for the `--print-tos` option. Please use something similar.

auerswal (changes requested):
This looks patterned after the `--print-tos` option, I like that. I also like that you added tests and documentation.

The code itself looks OK, but please address my comments regarding tests and man page.

ci/test-04-options-a-b.pl
nalves599: I thought that each command counted as a test. Thank you for your explanation! I changed to `41` (`32 + 9`).

ci/test-11-unpriv.pl
nalves599: Changed to `15`.

doc/fping.pod
nalves599: I did the same thing you did on that PR. Do you think it looks good?

ci/test-04-options-a-b.pl
```diff
@@ -68,6 +68,33 @@
 $cmd->stderr_is_eq("");
 }
 
+# fping -a --print-ttl
+{
+my $cmd = Test::Command->new(cmd => "fping -a --print-ttl 127.0.0.1 127.0.0.2");
+$cmd->exit_is_num(0);
+$cmd->stdout_like("127\.0\.0\.1 \(TTL \d+\)\n127\.0\.0\.2 \(TTL \d+\)\n");
```
auerswal: I think this needs to use `qr{…}`.

nalves599: Fixed :)

doc/fping.pod
```diff
@@ -140,6 +140,13 @@ Print usage message.
 
 Set the IP TTL field (time to live hops).
 
+=item B<--print-ttl>
+
+Displays the IPv4 TTL value from the IP Header in the output,
+that represents the hop limit. If B<fping> cannot read the TTL value,
```
auerswal: Please drop the ", that represents the hop limit" part.

auerswal (changes requested):
I hope that the tests now work as intended.
Please adjust the man page description (I mentioned hop count only for background, not as something that be added to the man page).

doc/fping.pod
auerswal: Please see my reply to your changed PR.

auerswal (approved):
I'd say this looks mostly OK.

The commit message could be a bit more explicit that this introduces a new option, but then your example commit has a similarly brief commit message.

I'll wait a bit before merging. This allows me to think more about the pull request, and gives others an opportunity to look at it, too.

PR #347, opened 2024-08-24
allow combination of -a with --print-tos

auerswal:
This is similar to how -e, --elapsed works.
Also adjust -h output regarding --print-tos.

(This conflicts with pull request #345 (I think), therefore only as a draft PR for now.)


gsnw-sebast:
The code is flawless and can be adopted. Tests have shown no abnormalities.

auerswal:
Thanks for the review, I'll just go ahead and merge.

PR #350, opened 2024-09-22
Dockerfile support architekture 386 change distribution from ubuntu to debian

gsnw-sebast:
The change of the Dockerfile from the distribution ubuntu to debian is for the support of the system architecture i386.
Successful test with podman `podman build -t fping -f contrib/Dockerfile .`


auerswal (approved):
The changes look fine (I do not use Docker, so please take this with a grain of salt).

Issue #265, opened 2022-09-20
Feature request: support ICMP type 13/14 timestamps


moeller0:
Sometimes for debugging it is helpful to get an idea about one way delays (or even just changes in one way delays). However the amount of servers willing to serve precise time (e.g. NTP servers) is quite low, not well distributed over the internet and often only willing to reply at low packet rates. However quite a number of publicly reachable servers are willing to respond to ICMP timestamps requests with the appropriate response. It would be excellent if fping could grow the capability to use ICMP timestamps instead of ICMP (caveat, ICMP over IPv6 tends not to support ICMP timestamps, but with large parts of the internet being IPv4 that still keeps ICMP timestamps relevant).

lynxthecat:
I would like to second this request. In addition to the excellent reasons provided above, I think this would serve to yet further increase the attractiveness of fping set against the other existing ping utilities which either: offer it but are buggy (e.g. nping); offer it but are no longer maintained (e.g. hping3); or do not even offer it ICMP type 13 (e.g. iputils-ping). ICMP type 13 requests remain a powerful tool for ascertaining not only RTT, but also its constituent one way delays between the client and the server.

gsnw-sebast:
Does anyone know a host that supports ICMP type 13/14 or does anyone know a way to test this with 127.0.0.1?

[1] Page-16: https://www.rfc-editor.org/rfc/rfc792.html

moeller0:
I believe that 9.9.9.9 supports ICMP timetamps:

```
user@mac~ % sudo hping3 9.9.9.9 --icmp --icmp-ts -c 1
Password:
HPING 9.9.9.9 (utun4 9.9.9.9): icmp mode set, 28 headers + 0 data bytes
len=40 ip=9.9.9.9 ttl=56 id=12218 icmp_seq=0 rtt=43.0 ms
ICMP timestamp: Originate=65378647 Receive=65378771 Transmit=65378771
ICMP timestamp RTT tsrtt=43


--- 9.9.9.9 hping statistic ---
1 packets tramitted, 1 packets received, 0% packet loss
round-trip min/avg/max = 43.0/43.0/43.0 ms
```

In addition my OpenWrt based router (192.168.42.1) also responds:
```
user@mac~ % sudo hping3 192.168.42.1 --icmp --icmp-ts -c 1
HPING 192.168.42.1 (en7 192.168.42.1): icmp mode set, 28 headers + 0 data bytes
[...]

PR #351, opened 2024-09-23
Add feature: support ICMP type 13/14 timestamps

gsnw-sebast:
This is a development branch. (Issues: #265)

The current version has no output yet, but can be tested with GDP. Please note that root rights are required for icmp type 13 queries

**Build**
```
./autogen.sh
./configure --enable-debug
make CFLAGS="-g -O0 -fprofile-arcs -ftest-coverage"
```

**GDB Usage**
```
gdb fping
set args --icmp-timestamp 127.0.0.1
break fping.c:2289

print *icp
```

### DIFF
--- a/ci/test-04-options-a-b.pl
+++ b/ci/test-04-options-a-b.pl
@@ -1,6 +1,6 @@
 #!/usr/bin/perl -w
 
-use Test::Command tests => 41;
+use Test::Command tests => 44;
 use Test::More;
 use Time::HiRes qw(gettimeofday tv_interval);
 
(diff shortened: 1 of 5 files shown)


gsnw-sebast:
The current update outputs the output directly with `./fping --icmp-timestamp 127.0.0.1`

I'm still thinking about putting the ip header returns from decode_icmp_ipv4 into an extra structure. Because the many returns make it a bit overloaded

gsnw-sebast:
Development is finished and therefore the pull-request is closed and a new one is created with the final version

PR #353, opened 2024-10-08
Add feature: support ICMP type 13/14 timestamps

gsnw-sebast:
This is an extension to send ICMP type 13 and receive ICMP type 14 with fping (Issue: #265).
To use it, root rights are required on the system and it can be used as follows.

1. `./fping --icmp-timestamp 127.0.0.1`
```
127.0.0.1 is alive (Timestamp Originate=67895434 Receive=67895434 Transmit=67895434)
```
2. `./fping --icmp-timestamp -c 2 127.0.0.1` (For the output, I first partly oriented myself on hping3)
```
27.0.0.1 : [0], 20 bytes, 0.072 ms (0.072 avg, 0% loss)
ICMP timestamp: Originate=67951889 Receive=67951889 Transmit=67951889
ICMP timestamp RTT tsrtt=71991
127.0.0.1 : [1], 20 bytes, 0.057 ms (0.064 avg, 0% loss)
ICMP timestamp: Originate=67952890 Receive=67952890 Transmit=67952890
ICMP timestamp RTT tsrtt=56783

127.0.0.1 : xmt/rcv/%loss = 2/2/0%, min/avg/max = 0.057/0.064/0.072
``` 

Unfortunately I could not test the function under macOS and the Github Action Test under macOS is not possible.

### DIFF
--- a/doc/fping.pod
+++ b/doc/fping.pod
@@ -157,8 +157,8 @@ Set the interface (requires SO_BINDTODEVICE support).
 
 =item B<--icmp-timestamp>
 
-Send ICMP timestamp requests (ICMP type 13).
-If B<fping> not recive ICMP timestamp reply, "(Timestamp unknown)" is returned.
+Send ICMP timestamp requests (ICMP type 13) instead of ICMP Echo requests.
+Cannot be used together with B<-b> because ICMP timestamp messages have a fixed size.
 IPv4 only, requires root privileges.
 
 =item B<-k>, B<--fwmark>=I<FWMARK>
--- a/doc/fping.pod
+++ b/doc/fping.pod
@@ -155,6 +155,12 @@ to any target (default is 10, minimum is 1).
 
 Set the interface (requires SO_BINDTODEVICE support).
 
+=item B<--icmp-timestamp>
+
+Send ICMP timestamp requests (ICMP type 13).
+If B<fping> not recive ICMP timestamp reply, "(Timestamp unknown)" is returned.
+IPv4 only, requires root privileges.
+
 =item B<-k>, B<--fwmark>=I<FWMARK>
 
 Set FWMARK on ping packets for policy-based routing. Requires Linux kernel
--- a/src/fping.c
+++ b/src/fping.c
@@ -1939,7 +1939,7 @@ int send_ping(HOST_ENTRY *h, int index)
 
     if (h->saddr.ss_family == AF_INET && socket4 >= 0) {
         if(icmp_request_typ == 13)
-            proto = 13;
+            proto = ICMP_TSTAMP;
         n = socket_sendto_ping_ipv4(socket4, (struct sockaddr *)&h->saddr, h->saddr_len, myseq, ident4, proto);
     }
 #ifdef IPV6
@@ -2216,7 +2216,7 @@ int decode_icmp_ipv4(
 
     icp = (struct icmp *)(reply_buf + hlen);
 
-    if (icp->icmp_type != ICMP_ECHOREPLY && icp->icmp_type != 14) {
+    if (icp->icmp_type != ICMP_ECHOREPLY && icp->icmp_type != ICMP_TSTAMPREPLY) {
         /* Handle other ICMP packets */
         struct icmp *sent_icmp;
         SEQMAP_VALUE *seqmap_value;
@@ -2231,7 +2231,7 @@ int decode_icmp_ipv4(
 
         sent_icmp = (struct icmp *)(reply_buf + hlen + ICMP_MINLEN + sizeof(struct ip));
 
-        if ((sent_icmp->icmp_type != ICMP_ECHO && sent_icmp->icmp_type != 13) || sent_icmp->icmp_id != ident4) {
+        if ((sent_icmp->icmp_type != ICMP_ECHO && sent_icmp->icmp_type != ICMP_TSTAMP) || sent_icmp->icmp_id != ident4) {
             /* not caused by us */
             return -1;
         }
@@ -2282,7 +2282,7 @@ int decode_icmp_ipv4(
 
     *id = icp->icmp_id;
     *seq = ntohs(icp->icmp_seq);
-    if(icp->icmp_type == 14) {
+    if(icp->icmp_type == ICMP_TSTAMPREPLY) {
         *ip_header_otime_ms = ntohl(icp->icmp_dun.id_ts.its_otime);
         *ip_header_rtime_ms = ntohl(icp->icmp_dun.id_ts.its_rtime);
         *ip_header_ttime_ms = ntohl(icp->icmp_dun.id_ts.its_ttime);
--- a/src/fping.c
+++ b/src/fping.c
@@ -691,6 +691,8 @@ int main(int argc, char **argv)
         case 'b':
             if (sscanf(optparse_state.optarg, "%u", &ping_data_size) != 1)
                 usage(1);
+            if (icmp_request_typ > 0)
+                usage(1);
 
             break;
 
@@ -3129,7 +3131,7 @@ void usage(int is_error)
     fprintf(out, "   -t, --timeout=MSEC individual target initial timeout (default: %.0f ms,\n", timeout / 1e6);
     fprintf(out, "                      except with -l/-c/-C, where it's the -p period up to 2000 ms)\n");
     fprintf(out, "       --check-source discard replies not from target address\n");
-    fprintf(out, "       --icmp-timestamp send ping type Timestamp Request\n");
+    fprintf(out, "       --icmp-timestamp send ping type Timestamp Request (only if no -b specified)\n");
     fprintf(out, "\n");
     fprintf(out, "Output options:\n");
     fprintf(out, "   -a, --alive        show targets that are alive\n");
--- a/src/fping.c
+++ b/src/fping.c
@@ -2404,7 +2404,6 @@ int wait_for_reply(int64_t wait_time)
     long ip_header_otime_ms = -1;
     long ip_header_rtime_ms = -1;
     long ip_header_ttime_ms = -1;
-    int tsrtt;
 
     /* Receive packet */
     result = receive_packet(wait_time, /* max. wait time, in ns */
@@ -2610,18 +2609,16 @@ int wait_for_reply(int64_t wait_time)
             fprintf(stderr, " [<- %s]", buf);
         }
 
-        printf("\n");
-
         if (icmp_request_typ == 13) {
             if(ip_header_otime_ms != -1 && ip_header_rtime_ms != -1 && ip_header_ttime_ms != -1) {
-                printf("ICMP timestamp: Originate=%lu Receive=%lu Transmit=%lu\n", ip_header_otime_ms, ip_header_rtime_ms, ip_header_ttime_ms);
-                tsrtt = (ip_header_ttime_ms - ip_header_otime_ms) + (this_reply - (ip_header_rtime_ms - ip_header_otime_ms));
-                printf("ICMP timestamp RTT tsrtt=%d\n", tsrtt);
+                printf(", ICMP timestamp: Originate=%lu Receive=%lu Transmit=%lu", ip_header_otime_ms, ip_header_rtime_ms, ip_header_ttime_ms);
             }
             else {
-                printf("ICMP timestamp: unknown\n");
+                printf(", ICMP timestamp: unknown");
             }
         }
+
+        printf("\n");
     }
 
     return 1;
--- a/src/fping.c
+++ b/src/fping.c
@@ -361,6 +361,7 @@ int timestamp_format_flag = 0;
 int random_data_flag = 0;
 int cumulative_stats_flag = 0;
 int check_source_flag = 0;
+int icmp_request_typ = 0;
 int print_tos_flag = 0;
 int print_ttl_flag = 0;
 #if defined(DEBUG) || defined(_DEBUG)
@@ -536,6 +537,7 @@ int main(int argc, char **argv)
         { "ttl", 'H', OPTPARSE_REQUIRED },
         { "interval", 'i', OPTPARSE_REQUIRED },
         { "iface", 'I', OPTPARSE_REQUIRED },
+        { "icmp-timestamp", '0', OPTPARSE_NONE },
 #ifdef SO_MARK
         { "fwmark", 'k', OPTPARSE_REQUIRED },
 #endif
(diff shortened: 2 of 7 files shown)


auerswal:
@gsnw-sebast: Thanks for your work on this!

I have not yet found the time to do a thorough review, but the approach looks sensible. I'll try to review this pull request in a couple of weeks.

In your example output for `--count=2`, you compute a *ICMP timestamp RTT* with a higher resolution than that if the ICMP timestamps. This is based on existing `fping` RTT measurements, I think. IMHO this is a bit misleading, since it does not use only the ICMP timestamp data (which has a millisecond resolution). What do you think?

Then I have two high-level suggestions for consideration:

1. Should `--icmp-timestamp` automatically set `--ipv4` since the ICMP timestamp and timestamp reply messages are only defined for IPv4?
2. Should the option combination `--icmp-timestamp --size=N` be rejected since the ICMP timestamp message has a fixed size?

gsnw-sebast:
I'm not sure about the RTT issue myself.
I have also tried to orientate myself on `hping3` here.

The formula should be `TSRTT = (Transmittime - Originatetime) + (ICMP RTT - (Transmittime - Originatetime)`
Here it could be that the ICMP RTT is not in `this_reply` line `2618`

```
if (icmp_request_typ == 13) {
            if(ip_header_otime_ms != -1 && ip_header_rtime_ms != -1 && ip_header_ttime_ms != -1) {
                printf("ICMP timestamp: Originate=%lu Receive=%lu Transmit=%lu\n", ip_header_otime_ms, ip_header_rtime_ms, ip_header_ttime_ms);
                tsrtt = (ip_header_ttime_ms - ip_header_otime_ms) + (this_reply - (ip_header_rtime_ms - ip_header_otime_ms));
                printf("ICMP timestamp RTT tsrtt=%d\n", tsrtt);
            }
            else {
                printf("ICMP timestamp: unknown\n");
            }
        }
```

1) According to my IPv6 book and the IANA page [1], ICMP Timestamp does not exist under IPv6
2) I would have to check that, thanks for the feedback (**Answer**: With ICMP Timestamp there is no data area in the header. It therefore has a fixed size and format)

[1] https://www.iana.org/assignments/icmpv6-parameters/icmpv6-parameters.xhtml

auerswal:
I would prefer `fping` not to compute an *ICMP timestamp RTT*. I would also prefer the ICMP timestamp values to always be in the same line as the other output for a received reply message. This could look as follows:

```
$ fping --icmp-timestamp -c 2 127.0.0.1
127.0.0.1 : [0], 20 bytes, 0.072 ms (0.072 avg, 0% loss), ICMP timestamp: Originate=67951889 Receive=67951889 Transmit=67951889
127.0.0.1 : [1], 20 bytes, 0.057 ms (0.064 avg, 0% loss), ICMP timestamp: Originate=67952890 Receive=67952890 Transmit=67952890

127.0.0.1 : xmt/rcv/%loss = 2/2/0%, min/avg/max = 0.057/0.064/0.072
```

gsnw-sebast:
The format you suggested was also my first idea. I have therefore adapted it. Thanks for the feedback.

gsnw-sebast:
I have changed the parameters so that --icmp-timestamp and --size may not be used at the same time.
Even if e.g. --icmp-timestamp works with e.g. --size in the appropriate size

I know the documentation is still missing

doc/fping.pod
```diff
@@ -155,6 +155,12 @@ to any target (default is 10, minimum is 1).
 
 Set the interface (requires SO_BINDTODEVICE support).
 
+=item B<--icmp-timestamp>
+
+Send ICMP timestamp requests (ICMP type 13).
+If B<fping> not recive ICMP timestamp reply, "(Timestamp unknown)" is returned.
```
auerswal: Please insert the word "does" before "not".

src/fping.c
```diff
@@ -2272,6 +2284,12 @@ int decode_icmp_ipv4(
 
     *id = icp->icmp_id;
     *seq = ntohs(icp->icmp_seq);
+    if(icp->icmp_type == ICMP_TSTAMPREPLY) {
```
auerswal, line 2287: I'd think that we need to check that `reply_buf_len` is sufficiently big to contain the timestamps. Otherwise a broken implementation (or an attacker) could send a short reply packet with correct header, but without timestamps, and cause a buffer overread by `fping`.

src/fping.c
```diff
@@ -2384,6 +2402,10 @@ int wait_for_reply(int64_t wait_time)
     unsigned short seq;
     int ip_header_tos = -1;
     int ip_header_ttl = -1;
+    // ICMP Timestamp
+    long ip_header_otime_ms = -1;
```
auerswal: I would prefer an exact length type, e.g., `int64_t` or `int32_t`. Or perhaps even `uint32_t` with an initialization to `0xffffffff`.

The timestamps are defined to be contained in a 32 bit field each. One day has 86,400,000ms (or 86,401,000 with a leap second). Values above cannot be valid, and only conform to RFC 792 when the highest bit is set. Setting the highest bit would result in a negative value when interpreting the 32 bit field as a signed integer. As such, a timestamp may contain a value of `-1`.

src/fping.c
```diff
@@ -2539,6 +2564,15 @@ int wait_for_reply(int64_t wait_time)
               }
             }
 
+            if (icmp_request_typ == 13) {
+                if(ip_header_otime_ms != -1 && ip_header_rtime_ms != -1 && ip_header_ttime_ms != -1) {
```
auerswal: I'd think that any 32 bit timestamp value should be accepted. Since ICMP Timestamp support requires raw sockets, `fping` can always read the timestamps, so we do not need to handle the case of not being able to read the data here. I'd prefer to always print the timestamp value.

src/fping.c
```diff
@@ -2577,6 +2611,15 @@ int wait_for_reply(int64_t wait_time)
             fprintf(stderr, " [<- %s]", buf);
         }
 
+        if (icmp_request_typ == 13) {
+            if(ip_header_otime_ms != -1 && ip_header_rtime_ms != -1 && ip_header_ttime_ms != -1) {
```
auerswal: Again, I'd prefer to always print the received timestamp value.

src/fping.c
```diff
@@ -3088,6 +3131,7 @@ void usage(int is_error)
     fprintf(out, "   -t, --timeout=MSEC individual target initial timeout (default: %.0f ms,\n", timeout / 1e6);
     fprintf(out, "                      except with -l/-c/-C, where it's the -p period up to 2000 ms)\n");
     fprintf(out, "       --check-source discard replies not from target address\n");
+    fprintf(out, "       --icmp-timestamp send ping type Timestamp Request (only if no -b specified)\n");
```
auerswal: Perhaps use the following?

`--icmp-timestamp use ICMP Timestamp instead of ICMP Echo`

I don't think that the incompatibility with setting a size needs to be mentioned in the (rather succinct) `--help` output.

doc/fping.pod
```diff
@@ -155,6 +155,12 @@ to any target (default is 10, minimum is 1).
 
 Set the interface (requires SO_BINDTODEVICE support).
 
+=item B<--icmp-timestamp>
+
+Send ICMP timestamp requests (ICMP type 13).
```
auerswal: Perhaps add the following to the end of the sentence?

` instead of ICMP Echo requests`

doc/fping.pod
```diff
@@ -155,6 +155,12 @@ to any target (default is 10, minimum is 1).
...
+If B<fping> not recive ICMP timestamp reply, "(Timestamp unknown)" is returned.
+IPv4 only, requires root privileges.
```
auerswal, line 162: Perhaps insert something like the following?

`Cannot be used together with B<-b> because ICMP timestamp messages have a fixed size.`

auerswal (changes requested):
All in all this looks promising, I think we just need to look at some of the details before merging. Thanks!

src/fping.c
gsnw-sebast: But then I wouldn't need the return `(Timestamp unknown)`, or I would have to check the return to see if `ICMP_TSTAMPREPLY` really happened.
If `ICMP_TSTAMP` is blocked by a firewall, either way only `is unreachable` appears

I don't know if there is a way to answer type 13 without type 14. Because that should be intercepted.

auerswal: An ICMP Timestamp is either answered with an ICMP Timestamp Reply, or possibly an error ICMP message (e.g., ICMP Destination Unreachable). When blocked by a firewall, no reply at all would be common. When no ICMP Timestamp functionality is offered, no reply at all would be expected, as far as I understand it.

Every ICMP Timestamp Reply must include the timestamps, thus I'd say that `(Timestamp unknown)` should never happen and is not really needed. If the packet is too short to contain the timestamps, it should be handled the same way as a too short ICMP Echo reply or other too short messages (see, e.g., lines 2197 and 2309 of `fping.c`).

doc/fping.pod
gsnw-sebast: Superfluous due to the omission of `(Timestamp unknown)`

src/fping.c
gsnw-sebast: I have add line  `2289`

```
        /* Check that reply_buf_len is sufficiently big to contain the timestamps */
        if (reply_buf_len < hlen + ICMP_MINLEN + 3 * sizeof(uint32_t)) {
            if (verbose_flag) {
                char buf[INET6_ADDRSTRLEN];
                getnameinfo(response_addr, response_addr_len, buf, INET6_ADDRSTRLEN, NULL, 0, NI_NUMERICHOST);
                printf("received packet too short for ICMP Timestamp Reply (%d bytes from %s)\n", (int)reply_buf_len, buf);
            }
            return -1;
        }
```

src/fping.c
gsnw-sebast: Has been changed to uint32_t as this is sufficient and 0xffffffffff is not required.

src/fping.c
gsnw-sebast: Removed the check

schweikert (approved):
This looks good, thank you!

Suggestion: could you also update CHANGELOG.md?

auerswal (approved):
Looks good, thanks!

gsnw-sebast:
> Suggestion: could you also update CHANGELOG.md?

Shouldn't the CHANGELOG.md be created again with release 5.3 and then the rule defined that every pull request must also contain the extension of the CHANGELOG.md?

schweikert:
> > Suggestion: could you also update CHANGELOG.md?
> 
> Shouldn't the CHANGELOG.md be created again with release 5.3 and then the rule defined that every pull request must also contain the extension of the CHANGELOG.md?

Ideally yes, but you could already start with adding a line to a "Next" section maybe? We definitely need to get the habit of always also changing CHANGELOG.md instead of me guessing a good name for the change summary based on the commits.

moeller0:
This is excellent! However one request, would it be possible to also add the missing 4th number, the time the type 14 packet was received in the same time format as Originate? The goal is to calculate the two OWDs:
Forward OWD: Receive - Originate
Return OWD: MISSINGLOCALRECEIVE - Transmit

and to do this MISSINGLOCALRECEIVE would be helpful, or in the hping case that can be constructed as:
Originate + tsrtt

I guess the same is possible for fping:
```
user@macbook src % sudo ./fping -c 1 --icmp-timestamp 9.9.9.9
Password:
9.9.9.9 : [0], 20 bytes, 40.9 ms (40.9 avg, 0% loss), ICMP timestamp: Originate=55736120 Receive=55736157 Transmit=55736157
```

MISSINGLOCALRECEIVE = Originate + 40.9

it is just parsing this would be easier if there was something like:
` Originate=55736120 Receive=55736157 Transmit=55736157 Localreceive=55736160`

lynxthecat:
Sweet, so fping is about to support ICMP type 13/14?

moeller0:
It does, at least the development version!

lynxthecat:
> This is excellent! However one request, would it be possible to also add the missing 4th number, the time the type 14 packet was received in the same time format as Originate? The goal is to calculate the two OWDs: Forward OWD: Receive - Originate Return OWD: MISSINGLOCALRECEIVE - Transmit
> 
> and to do this MISSINGLOCALRECEIVE would be helpful, or in the hping case that can be constructed as: Originate + tsrtt
> 
> I guess the same is possible for fping:
> 
> ```
> user@macbook src % sudo ./fping -c 1 --icmp-timestamp 9.9.9.9
> Password:
> 9.9.9.9 : [0], 20 bytes, 40.9 ms (40.9 avg, 0% loss), ICMP timestamp: Originate=55736120 Receive=55736157 Transmit=55736157
> ```
> 
> MISSINGLOCALRECEIVE = Originate + 40.9
> 
> it is just parsing this would be easier if there was something like: ` Originate=55736120 Receive=55736157 Transmit=55736157 Localreceive=55736160`

Was this implemented?

auerswal:
```
$ ./src/fping -v
./src/fping: Version 5.3
$ ./src/fping --icmp-timestamp localhost
localhost is alive, timestamps: Originate=42694656 Receive=42694656 Transmit=42694656 Localreceive=42694656
```

lynxthecat:
Any idea when this will make its way into master or an official release? I am mainly curious to know when it’ll make its way into OpenWrt’s fping package.

auerswal:
This is in the default branch (`develop`) since last December, see https://github.com/schweikert/fping/commit/a5124e35d07cc5b244287ea10c2a59a321b26fdf
This is included in fping relase 5.3 from January, see https://github.com/schweikert/fping/releases/tag/v5.3

gsnw-sebast:
For openwrt please contact the administrator of the fping package of openwrt, 5.2 is still used there. As auerswal already said, it is included in fping-5.3

lynxthecat:
Excellent. And thank you both! Very exciting that fping now supports ICMP type 13 requests. This unlocks the use of one way delays for both the upload and download direction (rather than simply RTT/2) for users of [cake-autorate](https://github.com/lynxthecat/cake-autorate) that rely on fping for the ping binary (this is the default).

lynxthecat:
@gsnw-sebast I have a question about this implementation. Is the RTT that’s printed out computed independently from the timestamps? There are situations when the timestamps based on local and remote clocks will give rise to incorrect OWDs unless corrected. Having independently computed RTTs would make it easier to identify these occurrences by simply comparing the sums of the OWDs determined using the timestamps and the RTTs determined independently therefrom.

auerswal:
The RTT printed with `--count`, `--vcount`, `--loop`, or `--elapsed` is computed from the local sending and receiving time. The ICMP Timestamp Reply data is only printed, not used for computations.

The printed RTT should always correspond with *Localreceive* - *Originate* (it may have more precision). It should always correspond to the sum of the one-way delays plus the remote computation time. **Edit**: The printed *Originate* value is taken from the received ICMP Timestamp Reply packet and may differ from the value created by `fping`.

The printed RTT, though computed independently of received timing data, does not provide extra information (except possibly precision) in addition to the four timestamps printed with `--icmp-timestamp`, because both *Originate* and *Localreceive* are determined locally. **Edit**: The printed RTT does provide a sanity check for the *Localreceive* - *Originate* RTT, because the printed *Originate* value comes from the network and thus cannot be trusted (in general).

PR #354, opened 2024-10-14
Add docker build to azure-pipline

gsnw-sebast:
This only test docker build under azure-pipline


auerswal:
I am not sure what exactly is intended by this pull request, or how it works, but it does not look problematic to me. At least for now, I will not merge it, but I do not oppose merging by someone else.

gsnw-sebast:
It is only one way to check with Azure DevOps whether the dockerfile works.
It could also be the basis for providing a finished docker image or for testing the build process under i386.
The whole thing could also be transferred to Github Action later.

I would do the merge myself in about 2 weeks if there are no objections by then.

schweikert (approved):

PR #359, opened 2024-11-15
Add autotrigger for changes in develop to the azure-pipeline

gsnw-sebast:
This change caused the azure-pipeline to start automatically only when changes were made in the branch develop.
For all other changes, the process must be started manually in azure-pipline.

In addition, CHANGELOG.md was extended for the last two pull requests to the azure-pipeline


auerswal:
This looks OK.

Issue #352, opened 2024-10-05
CI job "Release-Tarball" no longer works


auerswal:
After merging pull request #350, the automatically triggered CI job "Release-Tarball" failed with the error that it uses a deprecated API version:

> This request has been automatically failed because it uses a deprecated version of `actions/upload-artifact: v2`. Learn more: https://github.blog/changelog/2024-02-13-deprecation-notice-v1-and-v2-of-the-artifact-actions/

![fping-error](https://github.com/user-attachments/assets/3f8013d2-1156-49fe-b757-948d2a3cce3c)

gsnw-sebast:
Pull request #355 was created for this purpose. However, the question is whether the intermediate step should be taken via v3 and we then go to v4 or directly to v4.

Version v4 apparently requires uniqueness in the `name` variable, which can be achieved via `${{ matrix.os }}` in `name`.

**Untested**
```
- uses: actions/upload-artifact@v2
        with:
          name: dist-${{ matrix.os }}
          path: fping-*.tar.gz
```

auerswal:
I would suggest to merge PR #355 and see if it works as intended. Then we can look into using v4. Since you already started looking into it, feel free to create a pull request.

gsnw-sebast:
Version 4 works `name: dist-${{ github.sha }}` #360

auerswal:
If I understand everything correctly, the CI job already works again. We can leave this issue open until the change to version 4 is completed.

PR #360, opened 2024-11-15
ci: Upgrade actions/upload-artifact to v4

gsnw-sebast:
actions/upload-artifact@v4 works directly because there is already a distinction in the jobs
Issue: #352


auerswal:
This looks OK.

schweikert (approved):

PR #363, opened 2024-11-30
Type13 tweaks

### LINKED ISSUE #361: FR: add local receive timestamp to `--icmp-timestamp`
@moeller0 asked for the addition of a local receive timestamp value to the output from `--icmp-timestamp` in https://github.com/schweikert/fping/pull/353#issuecomment-2476739176. The existing `-D, --timestamp` option requires loop or count mode, and `--timestamp-format` does not provide the RFC 792 format of *milliseconds since midnight UTC*. Thus the output of `--icmp-timestamp` should be amended similarly to @moeller0's suggestion.

I too think that something like this would be useful for tracking *changes* in one way delay.

auerswal:
Some adjustments for ICMP Timestamp support. IMHO this does not need extra mention in `CHANGELOG.md`, maybe just add the pull request number to the line announcing ICMP Timestamp support.

Please see the individual commit messages for more information.

### DIFF
--- a/doc/fping.pod
+++ b/doc/fping.pod
@@ -51,7 +51,7 @@ timestamp).  The reported received data size includes the IP header (normally
 20 bytes) and ICMP header (8 bytes), so the minimum total size is 40 bytes.
 Default is 56, as in B<ping>. Maximum is the theoretical maximum IP datagram
 size (64K), though most systems limit this to a smaller, system-dependent
-number.
+number. Cannot be used together with B<--icmp-timestamp>.
 
 =item B<-B>, B<--backoff>=I<N>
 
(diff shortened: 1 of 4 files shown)


doc/fping.pod
gsnw-sebast, line 1: With cap_net_raw the execution of fping is allowed and there is no socket error but ICMP_TIMESTAMP only works with root. The error then occurs `127.0.0.1: error while sending ping: Invalid argument`

gsnw-sebast: Sorry, my mistake. My build script is broken at `setcap cap_net_raw,cap_net_admin+ep`

gsnw-sebast (approved):
No objections from my side. The changes are very good.

moeller0:
Excellent! Looks great.

PR #375, opened 2025-01-16
complete SO_TIMESTAMPNS to SO_TIMESTAMP fallback

### LINKED ISSUE #374: When falling back to SO_TIMESTAMP reply timestamp is probably not working
I noticed in this commit 
https://github.com/schweikert/fping/commit/e8660637ccc3bf2aae163fa7c26216843ae98180#commitcomment-151310711
 that when fping can't setsockopt SO_TIMESTAMPNS it falls back to SO_TIMESTAMP. But the reading of the cmsg data is not considering this fallback. In that case no timestamp would be read. I think the fallback is not completely implemented. In that case the reply time is still coming from this line https://github.com/schweikert/fping/blob/cb832861742f4af5fcadfaf6b2dd19ee0a4b4143/src/fping.c#L2454

Relevant documentation: https://docs.kernel.org/networking/timestamping.html

auerswal:
Commit e8660637ccc3bf2aae163fa7c26216843ae98180 added a fallback from setting the socket option SO_TIMESTAMPNS to setting the socket option SO_TIMESTAMP if the nanosecond timestamp option could not be set.  But it did not add code to also look for the control message related to SO_TIMESTAMP.  Thus microsecond timestamps were requested, but not read.

This commit adds the missing code to read microsecond timestamp control messages.

The problem was reported in GitHub issue #374 by @payload.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+Next
+====
+
+## Bugfixes and other changes
+
+- Fix fallback to SO\_TIMESTAMP if SO\_TIMESTAMPNS is not available (#375, thanks
+  @auerswal)
+
 fping 5.3 (2025-01-02)
 ======================
 
--- a/src/fping.c
+++ b/src/fping.c
@@ -1284,6 +1284,14 @@ static inline int64_t timespec_ns(struct timespec *a)
     return ((int64_t)a->tv_sec * 1000000000) + a->tv_nsec;
 }
 
+#if HAVE_SO_TIMESTAMPNS
+/* convert a struct timeval to nanoseconds */
+static inline int64_t timeval_ns(struct timeval *a)
+{
+    return ((int64_t)a->tv_sec * 1000000000) + ((int64_t)a->tv_usec * 1000);
+}
+#endif /* HAVE_SO_TIMESTAMPNS */
+
 void add_cidr(char *addr)
 {
     char *addr_end;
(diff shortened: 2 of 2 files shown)


gsnw-sebast:
Not sure, but the actual ticket had the problem, the binary was built with HAVE_SO_TIMESTAMPNS but the system itself could not use SO_TIMESTAMPNS. So not sure about the implications of timeval_ns() if it is still present in SCM_TIMESTAMP.

auerswal:
According to [man 7 socket](https://man7.org/linux/man-pages/man7/socket.7.html), the two socket options `SO_TIMESTAMP` and `SO_TIMESTAMPNS` result in the creation of different *control messages*:

* Setting `SO_TIMESTAMP` results in control messages of type `SCM_TIMESTAMP` with the data in `struct timeval` format for microsecond resolution.
* Setting `SO_TIMESTAMPNS` results in control messages of type `SCM_TIMESTAMPNS` with the data in `struct timespec` format for nanosecond resolution.

Since only one of the two socket options is set for each socket, only one type of control messages is created for packets received on the respective socket.

According to the [online man page](https://man7.org/linux/man-pages/man7/socket.7.html), "*A socket cannot mix `SO_TIMESTAMP` and `SO_TIMESTAMPNS`: the two modes are mutually exclusive.*"

I have tested this on GNU/Linux, where both `SO_TIMESTAMPNS` and `SO_TIMESTAMP` work, by adding debugging output to the `CMSG` related code and changing the preference of the two socket options:

* When `SO_TIMESTAMPNS` is tried first, this succeeds, and only `SCM_TIMESTAMPNS` type control messages are created.
* When `SO_TIMESTAMP` is tried first, this succeeds, and only `SCM_TIMESTAMP` type control messages are created.

I have then changed the code to prefer `SO_TIMESTAMP` for IPv4 and `SO_TIMESTAMPNS` for IPv6 and used a dual-stack target host: `fping -mAne target.example`. This showed `SCM_TIMESTAMP` for IPv4 and `SCM_TIMESTAMPNS` for IPv6, and similar elapsed times for both address families, as expected.

As such I think that the code in this pull request works as intended and is required to make use of the fallback to `SO_TIMESTAMP`.

gsnw-sebast (approved):
Thank you for your detailed reply. I have tested it and it looks good

PR #376, opened 2025-01-19
IPv6 support for -g, --generate

### LINKED ISSUE #59: Please allow -g at least for smaller IPv6 ranges
`fping6 -g` surely does not make sense in SLAAC environments or with other larger ranges at the size of /64.

But for DHCPv6 managed IPv6 ranges (like the /115 and /118 ranges we use at ETH) there is likely some use.

I have no idea where to make the cut for refusing or at least warning the user about pinging a too big range.

auerswal:


auerswal:
I have change the IPv6 generator code to **not** exclude the *Subnet-Router anycast address* when using CIDR format, and rebased the branch on top of `develop`.

schweikert (approved):
Nice work, thank you!

gsnw-sebast (approved):
I have tested it and it works without any complaints

auerswal:
Thanks for the reviews, I'll merge this shortly.

PR #380, opened 2025-02-12
Issue/164-alternate - JSON functionality

ghost:
This PR implements comprehensive JSON output support for fping as requested in #164, adding both compact and pretty-printed formatting options. Based on the wonderful code in #379 by @gsnw-sebast 😊

### New Features
1. **JSON Output Format**
   - New `-j/--json` flag for compact JSON output
   - New `--json-pretty` flag for formatted, human-readable output
   - Consistent JSON structure across all output types (per-system and global stats)

2. **Pretty Print Functionality**
   - Configurable indentation (default: 2 spaces)
   - Proper array and object formatting
   - Optimized spacing around key-value pairs
   - Clean handling of newlines

### Technical Implementation
- Added new source files:
  - `src/json.c`: Core JSON formatting implementation
  - `src/json.h`: JSON interface definitions

- Key Functions:
  - Enhanced `print_per_system_stats()` with JSON support
  - Enhanced `print_global_stats()` with JSON support
  - Added JSON utility functions for consistent formatting

### Example Output
**Pretty-printed format (`--json-pretty`):**
```json
{
  "hosts": [
    {
      "host": "1.1.1.1",
      "count": [
        "0.400",
        "0.565"
      ]
    }
  ]
}
```
### Compact format (--json):
```json
{"hosts":[{"host":"1.1.1.1","count":["0.400","0.565"]}]}
```

### Compatibility
 - No breaking changes
 - Default behavior remains unchanged
 - Existing output formats are preserved
 - JSON output is optional and controlled by flags

### Code Changes
 - Modified 4 files:
 - 172 insertions
 - 100 deletions
 - Primary changes in `src/fping.c`, `src/json.c`, and `src/json.h`

### Test Coverage Details

Added comprehensive test suite to verify JSON output functionality. The tests cover:

1. **Basic JSON Output** (`-J -c`)
   - Verifies correct host information
   - Validates statistics fields (xmt, rcv, loss)
   - Checks timing information (min, avg, max)
   - Ensures proper numeric formatting

2. **Pretty-Printed JSON** (`--json-pretty`)
   - Validates proper indentation structure
   - Verifies consistent spacing around key-value pairs
   - Ensures proper array and object formatting
   - Maintains same data structure as compact output

3. **Count-Based Statistics** (`-J -C`)
   - Tests array of ping responses
   - Verifies proper time formatting for each ping
   - Handles both successful pings and timeouts ("-")
   - Ensures correct array structure

4. **Multiple Host Support** (`-J -g`)
[...]

PR #386, opened 2025-04-12
New option -J / --json for the JSON output

### LINKED ISSUE #164: It would be cool if fping could return json with results

gsnw-sebast:
This version is based on pull-request #379 and #380 and on the information from the discussions.
The following new functions are provided for json and the parameter `-J` or `--json` always requires `-q` and `-c` or `-C`

* print_per_system_stats_json
* print_per_system_splits_json
* print_global_stats_json

Further areas would then be added in a subsequent version.

## Example Output

**fping -J -c 1 -q 127.0.0.1**
```
{"summary": {"host": "127.0.0.1", "xmt": 1, "rcv": 1, "loss": 0, "rttMin": 0.064, "rttAvg": 0.064, "rttMax": 0.064}}
```

**fping -J -C 1 -q 127.0.0.1**
```
{"vSum": {"host": "127.0.0.1", "values": [0.062]}}
```

**fping -J -C 1 -q 127.0.0.1 127.0.0.2**
```
{"vSum": {"host": "127.0.0.1", "values": [0.074]}}
{"vSum": {"host": "127.0.0.2", "values": [0.058]}}
```

**fping -J -C3 -p 100 -q 127.0.0.1**
```
{"vSum": {"host": "127.0.0.1", "values": [0.051, 0.052, 0.060]}}
```

**fping -J -c 1 -q -g 127.0.0.1/30**
```
{"summary": {"host": "127.0.0.1", "xmt": 1, "rcv": 1, "loss": 0, "rttMin": 0.103, "rttAvg": 0.103, "rttMax": 0.103}}
{"summary": {"host": "127.0.0.2", "xmt": 1, "rcv": 1, "loss": 0, "rttMin": 0.058, "rttAvg": 0.058, "rttMax": 0.058}}
```

### DIFF
--- a/src/fping.c
+++ b/src/fping.c
@@ -2123,7 +2123,7 @@ void print_per_system_stats_json(void)
                 if ((resp = h->resp_times[j]) >= 0)
                     fprintf(stdout, "%s", sprint_tm(resp));
                 else
-                    fprintf(stdout, "-");
+                    fprintf(stdout, "null");
             }
 
             fprintf(stdout, "]}");
--- a/src/fping.c
+++ b/src/fping.c
@@ -1012,7 +1012,7 @@ int main(int argc, char **argv)
     }
 
     if (json_flag && !count_flag) {
-        fprintf(stderr, "%s: with -J or --json required -c or -C\n", prog);
+        fprintf(stderr, "%s: option -J, --json requires -c or -C\n", prog);
         exit(1);
     }
 
--- a/src/fping.c
+++ b/src/fping.c
@@ -358,7 +358,7 @@ int64_t next_report_time; /* time next -Q report is expected */
 /* switches */
 int generate_flag = 0; /* flag for IP list generation */
 int verbose_flag, quiet_flag, stats_flag, unreachable_flag, alive_flag;
-int elapsed_flag, version_flag, count_flag, loop_flag, netdata_flag;
+int elapsed_flag, version_flag, count_flag, loop_flag, netdata_flag, json_flag;
 int per_recv_flag, report_all_rtts_flag, name_flag, addr_flag, backoff_flag, rdns_flag;
 int multif_flag, timeout_flag, fast_reachable;
 int outage_flag = 0;
@@ -392,10 +392,13 @@ int send_ping(HOST_ENTRY *h, int index);
 void usage(int);
 int wait_for_reply(int64_t);
 void print_per_system_stats(void);
+void print_per_system_stats_json(void);
 void print_per_system_splits(void);
+void print_per_system_splits_json(void);
 void stats_reset_interval(HOST_ENTRY *h);
 void print_netdata(void);
 void print_global_stats(void);
+void print_global_stats_json(void);
 void main_loop();
 void signal_handler(int);
 void finish();
@@ -552,6 +555,7 @@ int main(int argc, char **argv)
         { "ttl", 'H', OPTPARSE_REQUIRED },
         { "interval", 'i', OPTPARSE_REQUIRED },
         { "iface", 'I', OPTPARSE_REQUIRED },
+        { "json", 'J', OPTPARSE_OPTIONAL },
         { "icmp-timestamp", '0', OPTPARSE_NONE },
 #ifdef SO_MARK
         { "fwmark", 'k', OPTPARSE_REQUIRED },
@@ -933,6 +937,10 @@ int main(int argc, char **argv)
 #endif
             break;
 
+        case 'J':
+            json_flag = 1;
+            break;
+
         case 'T':
             /* This option is ignored for compatibility reasons ("select timeout" is not meaningful anymore) */
             break;
@@ -1000,6 +1008,11 @@ int main(int argc, char **argv)
 
     if (count_flag && loop_flag) {
         fprintf(stderr, "%s: specify only one of c, l\n", prog);
+        exit(1);
+    }
+
+    if (json_flag && !count_flag) {
+        fprintf(stderr, "%s: with -J or --json required -c or -C\n", prog);
         exit(1);
     }
 
@@ -1125,6 +1138,8 @@ int main(int argc, char **argv)
             fprintf(stderr, "  outage_flag set\n");
         if (netdata_flag)
             fprintf(stderr, "  netdata_flag set\n");
+        if (json_flag)
+            fprintf(stderr, "  json_flag set\n");
     }
 #endif /* DEBUG || _DEBUG */
 
@@ -1697,25 +1712,58 @@ void main_loop()
             stats_add(h, event->ping_index, 0, -1);
 
             if (per_recv_flag) {
+                if (json_flag)
+                    printf("{\"resp\": {");
+
                 if (timestamp_flag) {
-                    print_timestamp_format(current_time_ns, timestamp_format_flag);
+                    if (json_flag)
+                        printf("\"timestamp\": %.5f, ", (double)current_time_ns / 1e9);
+                    else
+                        print_timestamp_format(current_time_ns, timestamp_format_flag);
                 }
-                printf("%-*s : [%d], timed out",
-                    max_hostname_len, h->host, event->ping_index);
-                if (h->num_recv > 0) {
-                    printf(" (%s avg, ", sprint_tm(h->total_time / h->num_recv));
+                if (json_flag)
+                {
+                  printf("\"host\": \"%s\", ", h->host);
+                  printf("\"seq\": %d", event->ping_index);
                 }
                 else {
-                    printf(" (NaN avg, ");
+                    printf("%-*s : [%d], timed out",
+                        max_hostname_len, h->host, event->ping_index);
+                }
+                if (h->num_recv > 0) {
+                    if (json_flag)
+                        printf(", \"rttAvg\": %s", sprint_tm(h->total_time / h->num_recv));
+                    else
+                        printf(" (%s avg, ", sprint_tm(h->total_time / h->num_recv));
+                }
+                else {
+                    if (json_flag)
+                        printf(", \"rttAvg\": \"NaN\"");
+                    else
+                        printf(" (NaN avg, ");
                 }
                 if (h->num_recv <= h->num_sent) {
-                    printf("%d%% loss)",
-                        ((h->num_sent - h->num_recv) * 100) / h->num_sent);
+                    if (json_flag) {
+                        printf(", \"loss\": %d", ((h->num_sent - h->num_recv) * 100) / h->num_sent);
+                    }
+                    else {
+                        printf("%d%% loss)",
+                            ((h->num_sent - h->num_recv) * 100) / h->num_sent);
+                    }
                 }
                 else {
-                    printf("%d%% return)",
-                        (h->num_recv_total * 100) / h->num_sent);
+                    if (json_flag) {
+                        printf(", \"return\": %d", (h->num_recv_total * 100) / h->num_sent);
+                    }
+                    else {
+                        printf("%d%% return)",
+                            (h->num_recv_total * 100) / h->num_sent);
+                    }
                 }
+
+                if (json_flag)
+                    printf("}}");
+                
                 printf("\n");
             }
 
@@ -1826,15 +1874,23 @@ void main_loop()
 
         if (status_snapshot) {
             status_snapshot = 0;
-            print_per_system_splits();
+            if (json_flag)
+                print_per_system_splits_json();
+            else
+                print_per_system_splits();
         }
 
         /* Print report */
         if (report_interval && (loop_flag || count_flag) && (current_time_ns >= next_report_time)) {
-            if (netdata_flag)
+            if (netdata_flag) {
                 print_netdata();
-            else
+            }
+            else if (json_flag) {
+                print_per_system_splits_json();
+            }
+            else {
                 print_per_system_splits();
+            }
 
             while (current_time_ns >= next_report_time) {
                 next_report_time += report_interval;
@@ -1923,15 +1979,27 @@ void finish()
         }
     }
 
-    if (count_flag || loop_flag)
-        print_per_system_stats();
+    if (count_flag || loop_flag) {
+        if (json_flag)
+            print_per_system_stats_json();
+        else
+            print_per_system_stats();
+    }
 #if defined(DEBUG) || defined(_DEBUG)
-    else if (print_per_system_flag)
-        print_per_system_stats();
+    else if (print_per_system_flag) {
+        if (json_flag)
+            print_per_system_stats_json();
+        else
+            print_per_system_stats();
+    }
 #endif /* DEBUG || _DEBUG */
 
-    if (stats_flag)
-        print_global_stats();
+    if (stats_flag) {
+        if (json_flag)
+            print_global_stats_json();
+        else
+            print_global_stats();
+    }
 
     if (min_reachable) {
         if ((num_hosts - num_unreachable) >= min_reachable) {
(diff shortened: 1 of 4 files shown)


gsnw-sebast:
The command `fping -J -s -c 1 -q 127.0.0.1` is currently broken because the format of `print_global_stats_json` is still wrong

**fping -J -s -c 1 -q 127.0.0.1**
```
{"summary": {"host": "127.0.0.1", "xmt": 1, "rcv": 1, "loss": 0, "rttMin": 0.066, "rttAvg": 0.066, "rttMax": 0.066}}
}
"targets":       1, "alive":       1, "unreachable":       0, "unknown addresses":       0, "timeouts (waiting for response)":       0, "ICMP Echos sent":       1, "ICMP Echo Replies received":       1, "other ICMP received":       0, "ms (min round trip time)": 0.066, "ms (avg round trip time)": 0.066, "ms (max round trip time)": 0.066, "sec (elapsed real time)":
```

The question is also whether anyone has any suggestions for shortening the keys, I couldn't find any reference to this in the discussion.

gsnw-sebast:
Format fix `fping -J -s -c 1 -q 127.0.0.1`

```
{"summary": {"host": "127.0.0.1", "xmt": 1, "rcv": 1, "loss": 0, "rttMin": 0.064, "rttAvg": 0.064, "rttMax": 0.064}}
{"targets": 1, "alive": 1, "unreachable": 0, "unknown addresses": 0, "timeouts (waiting for response)": 0, "ICMP Echos sent": 1, "ICMP Echo Replies received": 1, "other ICMP received": 0, "ms (min round trip time)": 0.064, "ms (avg round trip time)": 0.064, "ms (max round trip time)": 0.064, "sec (elapsed real time)": 0.000}
```

gsnw-sebast:
Preliminary info the next version for “fping --json -c 2 8.8.8.8” gives the following output.

```
{"resp": {"host": "8.8.8.8", "seq": 0, "size": 64, "rtt": 10.4, "rttAvg": 10.4, "loss": 0}}
{"resp": {"host": "8.8.8.8", "seq": 1, "size": 64, "rtt": 10.0, "rttAvg": 10.2, "loss": 0}}
{"summary": {"host": "8.8.8.8", "xmt": 2, "rcv": 2, "loss": 0, "rttMin": 10.0, "rttAvg": 10.2, "rttMax": 10.4}}
```

gsnw-sebast:
@JoshIPT you can review the new version and sorry that I destroy the most of your changes.

src/fping.c
```diff
@@ -1981,6 +2016,80 @@ void print_per_system_stats(void)
...
+
+    for (i = 0; i < num_hosts; i++) {
+        h = table[i];
+
+        if (report_all_rtts_flag)
+            fprintf(stdout, "{\"vSum\": {");
+        else
+            fprintf(stdout, "{\"summary\": {");
+
+        fprintf(stdout, "\"host\": \"%-*s\", ", max_hostname_len, h->host);
```
auerswal: This adds spaces to shorter hostnames, e.g.:
```
$ ./src/fping -c1 -J 127.0.0.1 127.255.255.254
{"resp": {"host": "127.0.0.1      ", "seq": 0, "size": 64, "rtt": 0.030, "rttAvg": 0.030, "loss": 0}}
{"resp": {"host": "127.255.255.254", "seq": 0, "size": 64, "rtt": 0.026, "rttAvg": 0.026, "loss": 0}}
{"summary": {"host": "127.0.0.1      ", "xmt": 1, "rcv": 1, "loss": 0, "rttMin": 0.030, "rttAvg": 0.030, "rttMax": 0.030}}
{"summary": {"host": "127.255.255.254", "xmt": 1, "rcv": 1, "loss": 0, "rttMin": 0.026, "rttAvg": 0.026, "rttMax": 0.026}}
```
This means that the value of the key `"host"` needs to be post-processed to get at the actual hostname. I'd suggest not to add any spaces for the JSON output. (For the non-JSON output, this padding is useful to align the `:` in the output. I would not align the JSON output, but if you do want to align it, the padding needs to be outside the value, i.e., after the doublequote (`"`).

src/fping.c
```diff
@@ -1981,6 +2016,80 @@ void print_per_system_stats(void)
...
+                    fprintf(stdout, "\"outage(ms)\": %d, ", outage_ms);
+                }
+            }
+            else {
+                fprintf(stdout, "\"xmt\": %d, ", h->num_sent);
+                fprintf(stdout, "\"rcv\": %d, ", h->num_recv);
+                fprintf(stdout, "\"return\": %d, ", h->num_sent > 0 ? ((h->num_recv * 100) / h->num_sent) : 0);
+            }
+
+            if (h->num_recv) {
```
auerswal, line 2146: If no reply is received, the statistics output is not valid JSON, because there is a trailing comma:
```
$ ./src/fping -c1 -J 192.0.2.47
192.0.2.47 : [0], timed out (NaN avg, 100% loss)
{"summary": {"host": "192.0.2.47", "xmt": 1, "rcv": 0, "loss": 100, }}
$ ./src/fping -c1 -J 192.0.2.47 | grep summary | jq .
parse error: Expected another key-value pair at line 1, column 69
```

src/fping.c
```diff
@@ -2115,6 +2224,63 @@ void print_per_system_splits(void)
...
+        }
+
+        if (h->num_recv_i) {
+            avg = h->total_time_i / h->num_recv_i;
+            fprintf(stdout, ", \"rttMin\": %s, ", sprint_tm(h->min_reply_i));
+            fprintf(stdout, "\"rttAvg\": %s, ", sprint_tm(avg));
+            fprintf(stdout, "\"rttMax\": %s", sprint_tm(h->max_reply_i));
+        }
+
+        fprintf(stdout, "}\n");
```
auerswal: There is one closing curly bracket missing in the `intSum` output:
```
$ ./src/fping -c2 -Q1 -J 127.0.0.1
{"intSum": {"time": 1745082391,"host": "127.0.0.1", "xmt": 1, "rcv": 1, "loss": 0, "rttMin": 0.038, "rttAvg": 0.038, "rttMax": 0.038}
{"summary": {"host": "127.0.0.1", "xmt": 2, "rcv": 2, "loss": 0, "rttMin": 0.038, "rttAvg": 0.045, "rttMax": 0.053}}
$ ./src/fping -c2 -Q1 -J 127.0.0.1 | jq .
parse error: Expected separator between values at line 2, column 1
```

auerswal (changes requested):
Timeout events are not yet included in the JSON output:
```
$ ./src/fping -c2 -J 192.0.2.47 
192.0.2.47 : [0], timed out (NaN avg, 100% loss)
192.0.2.47 : [1], timed out (NaN avg, 100% loss)
{"summary": {"host": "192.0.2.47", "xmt": 2, "rcv": 0, "loss": 100, }}
```

src/fping.c
gsnw-sebast: Ah thanks, I've corrected that in the new version. It's a copy&paste error

src/fping.c
gsnw-sebast: I have extended the missing JSON output

```
{"resp": {"host": "192.0.2.47", "seq": 0, "rttAvg": "NaN", "loss": 100}}
{"summary": {"host": "192.0.2.47", "xmt": 1, "rcv": 0, "loss": 100}}
```

I could still include the time out text as `“error”: “time out”`

src/fping.c
gsnw-sebast: Fix in https://github.com/schweikert/fping/commit/9e4a90dfd0dec0ecdb3a1035068fad9ecb5f88ba

src/fping.c
gsnw-sebast: Fix in https://github.com/schweikert/fping/commit/9e4a90dfd0dec0ecdb3a1035068fad9ecb5f88ba

gsnw-sebast:
The following open points are still included where no JSON format is currently output

- [x] timestamp_flag
- [x] icmp_request_typ == 13
- [x] print_tos_flag
- [x] print_ttl_flag

src/fping.c
```diff
@@ -2924,6 +3204,13 @@ void add_name(char *name)
         if (!quiet_flag)
             print_warning("%s: %s\n", name, gai_strerror(ret_ga));
         num_noaddress++;
+
+        // Handle JSON output for invalid hosts
+        if (json_flag) {
+            fprintf(stderr, "{\"host\": \"%s\", \"error\": \"%s\"}", name, gai_strerror(ret_ga));
```
schweikert: I would always use stdout for the JSON output. Also, this line is not following the general structure:

{"log-type": { "field1": "...", "field2": "..." }}

src/fping.c
```diff
@@ -2818,21 +3056,39 @@ int wait_for_reply(int64_t wait_time)
 
     /* print received ping (unless --quiet) */
     if (per_recv_flag) {
+        if (json_flag)
```
schweikert, line 3110: I am perfectly fine doing this in a separate PR, but I am wondering whether some minimal abstraction could be implemented to make the code less messy. Something like:

```
if (per_recv_flag) {
   if(json_flag) {
     print_recv_json(recv_time, h, result, this_count, this_reply, avg)
   }
  else {
    print_recv(...)
  }
}
```

schweikert (commented):
Thanks a lot for this! Let get this merged as soon as possible and address any non-essential issues in further PRs.

Something that we should also do at some point is document the format (either in the man-page or in a separate markdown file).

src/fping.c
gsnw-sebast: You mean `fprintf(stdout, "{\"log-type\": {\"host\": \"%s\", \"error\": \"%s\"}}", name, gai_strerror(ret_ga));`

src/fping.c
gsnw-sebast: That's right, it would make sense and is hardly any different to the two functions `print_per_system_stats` and `print_per_system_stats_json()` only for recv

src/fping.c
schweikert: What I mean is that every emitted JSON line should follow the following structure:

```
{"LOGTYPE": "{"FIELD1": VALUE1, "FIELD2": VALUE2", etc.}
```

So for errors possibly something like:

```
{"warning": { "host": ..., "error": ...}
```


... or maybe for now just leave the non-JSON formatted error on stderr.

gsnw-sebast: Sorry for the confusion. I have changed it to the following and may add a separate function for outputting errors in JSON format later

fprintf(stdout, "{\"warning\": {\"host\": \"%s\", \"message\": \"%s\"}}", name, gai_strerror(ret_ga));

src/fping.c
auerswal: Thanks for fixing the trailing comma and adding a JSON version of the timeout message!
I think it would be better to call this a "timeout" event instead of "resp", because there was no response:

`{"timeout": {"host": "192.0.2.47", "seq": 0, "rttAvg": "NaN", "loss": 100}}`

src/fping.c
```diff
@@ -970,6 +978,11 @@ int main(int argc, char **argv)
         exit(1);
     }
 
+    if (json_flag && !count_flag) {
+        fprintf(stderr, "%s: with -J or --json required -c or -C\n", prog);
+        exit(1);
+    }
+
```
auerswal, line 1018: How about adding another check for options that are either ignored with `-J` or produce non-JSON output despite `-J`, and exit if this is detected?

src/fping.c
```diff
@@ -970,6 +978,11 @@ int main(int argc, char **argv)
         exit(1);
     }
 
+    if (json_flag && !count_flag) {
+        fprintf(stderr, "%s: with -J or --json required -c or -C\n", prog);
```
auerswal: How about the following error message formulation?

`"%s: option -J, --json requires -c or -C\n"`

src/fping.c
gsnw-sebast: Hmmm, I could have thought of that myself. Done

src/fping.c
```diff
@@ -1981,6 +2049,80 @@ void print_per_system_stats(void)
...
+        if (report_all_rtts_flag) {
+            fprintf(stdout, "\"values\": [");
+            for (j = 0; j < h->num_sent; j++) {
+                if (j > 0)
+                  fprintf(stdout, ", ");
+                
+                if ((resp = h->resp_times[j]) >= 0)
+                    fprintf(stdout, "%s", sprint_tm(resp));
+                else
+                    fprintf(stdout, "-");
```
auerswal: This prints `-`, but that is not a valid JSON value. A valid value would be `"-"`, or possibly `null`.

```
$ ./src/fping -J -C1 192.0.2.47 
{"timeout": {"host": "192.0.2.47", "seq": 0, "rttAvg": "NaN", "loss": 100}}
{"vSum": {"host": "192.0.2.47", "values": [-]}}
$ ./src/fping -J -C1 192.0.2.47 | jq -c .
{"timeout":{"host":"192.0.2.47","seq":0,"rttAvg":"NaN","loss":100}}
parse error: Invalid numeric literal at line 2, column 45
```

src/fping.c
```diff
@@ -2845,33 +3101,57 @@ int wait_for_reply(int64_t wait_time)
...
-                printf(" (TOS %d)", ip_header_tos);
+                if (json_flag)
+                    printf(", \"tos\": %d", ip_header_tos);
+                else
+                    printf(" (TOS %d)", ip_header_tos);
             }
             else {
-                printf(" (TOS unknown)");
+                if (json_flag)
+                    printf(", \"tos\": -1");
```
auerswal, line 3176: I would prefer the same output value for JSON and normal output, i.e. `"unknown"` instead of `-1`.

src/fping.c
```diff
@@ -2845,33 +3101,57 @@ int wait_for_reply(int64_t wait_time)
...
-              printf(" (TTL %d)", ip_header_ttl);
+              if (json_flag)
+                  printf(", \"ttl\": %d", ip_header_ttl);
+              else
+                  printf(" (TTL %d)", ip_header_ttl);
           }
           else {
-              printf(" (TTL unknown)");
+              if (json_flag)
+                  printf(", \"ttl\": -1");
```
auerswal, line 3191: I would prefer the same output value for JSON and normal output, i.e. `"unknown"` instead of `-1`.

src/fping.c
auerswal: I think this is just `--timestamp-format` and `-e, --elapsed` (both are silently ignored). For `-e`, this could just be documented. For `--timestamp-format`, I'd say JSON output could in general support the different formats, and I'd prefer `fping` to reject combining the `--json` and `--timestamp-format` options for now.

auerswal (changes requested):
Thanks, this is making progress. :-)

src/fping.c
gsnw-sebast: But then the output of TOS must or should always be a string, although it is actually a number. I didn't want to mix that.

auerswal: OK, that is a valid concern. I do think that different value types could be OK, but that depends on the application using the data, or the programming language used. Perhaps this key/value pair should just be omitted when it cannot be read? (Same for TTL, of course.) @schweikert: What do you think?

src/fping.c
gsnw-sebast: I will implement this in a new PR after all. I've already started, but the changes are quite extensive

gsnw-sebast:
@auerswal I have rewritten the --timestamp-format a bit for the JSON output. But I'm not sure whether I haven't built in another security problem.

If not and this is so good, I would extend the tests even further

src/fping.c
gsnw-sebast: I have rewritten the --timestamp-format a bit for the JSON output. But I'm not sure whether I haven't built in another security problem.

If not and this is so good, I would extend the tests even further

src/fping.c
gsnw-sebast: It's not really a mistake, depending on the RFC. But I changed it to zero because it is valid in every RFC

auerswal (commented):
I'd prefer to change the `print_timestamp_format` function to look at the global `json_flag` and adjust the format string accordingly. This would avoid allocating and returning another buffer.

src/fping.c
```diff
@@ -550,6 +553,7 @@ int main(int argc, char **argv)
         { "ttl", 'H', OPTPARSE_REQUIRED },
         { "interval", 'i', OPTPARSE_REQUIRED },
         { "iface", 'I', OPTPARSE_REQUIRED },
+        { "json", 'J', OPTPARSE_OPTIONAL },
```
auerswal: Please change `OPTPARSE_OPTIONAL` to `OPTPARSE_NONE`, because this option does not take any arguments.

gsnw-sebast: Changed

src/fping.c
gsnw-sebast: But the behavior with -e is identical to without JSON output and because the JSON output requires -c or -C, this fits from my point of view

src/fping.c
gsnw-sebast: Ah yes, that's better. Thank you

src/fping.c
auerswal: You are right, `-e` is silently ignored for `-cN` and `-CN`, so it should still be silently ignored when adding `-J`.

src/fping.c
auerswal: Or `null`, see also discussion comment https://github.com/schweikert/fping/discussions/381#discussioncomment-12961106

auerswal:
Thanks for the adjustments, @gsnw-sebast!

I'd say that there is just one important open issue before this could be merged: how to handle missing values (i.e., values that cannot be read (TOS or TTL with datagram socket or IPv6), or that cannot be computed (e.g., because of division by zero)). Changing the data format after merging is always more problematic than doing so beforehand.

I will not block merging on this issue forever, but I would really like to read another opinion on this matter, ideally by someone intending to actually use JSON output from `fping` (see also my respective discussion comment https://github.com/schweikert/fping/discussions/381#discussioncomment-12961106).

src/fping.c
auerswal: Well, since I seem the only one to complain, but I do not plan to use `--json`, I retract this objection.

src/fping.c
auerswal: Well, since I seem the only one to complain, but I do not plan to use `--json`, I retract this objection.

auerswal (commented):
I no longer request changes.

auerswal (approved):

auerswal:
@gsnw-sebast, I'd think that you can merge this if you like. As @schweikert wrote in https://github.com/schweikert/fping/pull/386#pullrequestreview-2792383057, we can address potential future issues when they arise.

JoshFinlayAU:
I am @JoshIPT and I'd like to apologize for abandoning these efforts after submitting my original pull request. I had a change in work situations and thus affected my ability to contribute. I do appreciate being credited on this release and would be happy to contribute to future works. Again, apologies.

PR #391, opened 2025-05-04
fix ICMPv6 error message handling and reporting

### LINKED ISSUE #390: Missing output of ICMPv6 error messages
For IPv4, `fping` prints information about received ICMP error messages:

    $ fping --ttl 1 -An -r0 -4 dns.google
    ICMP Time Exceeded from 172.20.10.1 for ICMP Echo sent to dns.google (8.8.4.4)
    dns.google (8.8.4.4) is unreachable

But this does not happen for IPv6:

    $ fping --ttl 1 -An -r0 -6 dns.google
    dns.google (2001:4860:4860::8844) is unreachable

The receipt of ICMPv6 error messages can be seen using, e.g., `tcpdump` (some GUA addresses are edited):

    $ fping --ttl 1 -An -r0 -6 dns.google
    17:48:34.765445 IP6 (flowlabel 0x3149a, hlim 1, next-header ICMPv6 (58) payload length: 64) 2001:db8::c:1 > 2001:4860:4860::8844: [icmp6 sum ok] ICMP6, echo request, seq 0
    17:48:34.766351 IP6 (hlim 64, next-header ICMPv6 (58) payload length: 112) 2001:db8::1 > 2001:db8::c:1: [icmp6 sum ok] ICMP6, time exceeded in-transit for 2001:4860:4860::8844
    dns.google (2001:4860:4860::8844) is unreachable

I have already started to look at the problem. There are at least two problems in the code:

1. Offset calculation for the original packet inside the ICMPv6 error message uses the IPv4 header size, and thus the code uses the wrong data to compare with the ICMP Echo identification value used by `fping`.
2. `fping` would use IPv4 ICMP names to report received ICMPv6 error messages, but the ICMPv6 codes and types for IPv6 differ from ICMP for IPv4.

auerswal:
Handling of received ICMPv6 messages different from Echo Reply was based on IPv4 code.  It would thus use a too small minimum size, and interpret the wrong data as ICMPv6 fields.  Thus received ICMPv6 error messages were either not reported at all, or were reported incorrectly.

This commit introduces the following changes to fix this and thus address GitHub issue #390:

- Filter out irrelevant ICMPv6 messages: fping needs to receive ICMPv6 Echo Reply and ICMPv6 error messages, nothing else (only relevant for raw sockets).
- Fix offset calculation to look at the included packet that caused this ICMPv6 error message.
- Ignore invoking packets that do not have ICMPv6 as the first IPv6 Next Header in the packet (there could be, e.g., IPv6 Extension Headers).
- Fix ICMPv6 message type determination, and print respective warning messages.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,8 @@ Next
 - Options --print-tos and --print-ttl now also work for IPv6, and no
   longer require privileges (#384, thanks @auerswal)
 
+- Report received ICMPv6 error messages (#391, thanks @auerswal)
+
 fping 5.3 (2025-01-02)
 ======================
 
--- a/configure.ac
+++ b/configure.ac
@@ -34,7 +34,14 @@ AC_ARG_ENABLE([ipv6],
   AS_HELP_STRING([--disable-ipv6], [Disable support for pinging IPv6 hosts]))
 AS_IF([test "x$enable_ipv6" != "xno"], [
     dnl Test if IPv6 is supported
-       AC_CHECK_HEADERS([netinet/icmp6.h], [have_ipv6="yes"], [], [[
+       AC_CHECK_HEADERS([netinet/icmp6.h], [have_ipv6="yes"], [have_ipv6="no"], [[
+      #include <netinet/in.h>
+      #include <sys/types.h>
+  ]])
+])
+AS_IF([test "x$have_ipv6" == "xyes"], [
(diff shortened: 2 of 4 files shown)


gsnw-sebast:
I'm still testing

gsnw-sebast (approved):
First of all, all standard tests are run. The code also looks good. Unfortunately, producing certain errors is not so easy.

PR #392, opened 2025-05-04
suppress duplicate reports with -c/-C and -q/-Q

### LINKED ISSUE #389: How to prevent fping from displaying duplicate response information？
I use the command to check network connectivity, for example "fping -4 -b 56 -R -M -O 0x0 -c 20 -p 50 -t 500 www.google.com -q",  but sometimes it returns duplicate response results, such as "36.152.44.93 : duplicate for [16], 64 bytes, 119 ms". 

I don't want fping to display duplicate response results, what can I do?

auerswal:
If fping receives a duplicate Echo Reply packet, it reports this.  This report was suppressed for loop mode, and for other operating modes where each receive event is reported. Now, this duplicate report is also suppressed when in count mode with options -q, --quiet or -Q, --squiet.

This addresses GitHub issue #389.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,6 @@ Next
   @JoshIPT, @sebast-gsnw and @auerswal)
 - The -g, --generate option now also supports IPv6 addresses (#376,
   thanks @auerswal)
-
 - New option --seqmap-timeout to control the time after which sequence numbers
   can be used again (#388, thanks @auerswal)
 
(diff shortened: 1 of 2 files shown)


auerswal:
I have created this as a draft, because I am not sure if there was a reason to treat loop mode and count mode differently regarding duplicate reporting.

auerswal:
Any reasons against aligning duplicate report suppression in *count mode* with that of *loop mode*?

gsnw-sebast (approved):
From my point of view, we can take it from there.

auerswal:
While rebasing, I noticed that I had introduced blank lines between changelog items in a block. This was inconsistent with older changelog entries, so I removed those blank lines while solving the `CHANGELOG.md` merge conflict. I hope that is OK.

PR #394, opened 2025-05-17
Fix: OpenBSD sprintf warning

gsnw-sebast:
This fixes the message `sprintf() is often misused, please use snprintf()` under OpenBSD.
In general, the use of snprintf() is preferable to sprint() because it limits the number of characters and can thus actively prevent a buffer overflow


auerswal (approved):
This looks good, thanks. :-)

PR #400, opened 2025-07-28
Print receive ping moved to new functions

gsnw-sebast:
I have split the output below the function `int wait_for_reply(int64_t wait_time)` into four new functions.

* print_recv
* print_recv_json
* print_recv_ext
* print_recv_ext_json

This should make it easier to read.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ Next
   (#392, thanks @gsnw-sebast and @auerswal)
 - Switch to alpine-based multi-stage Docker build to reduce image size and improve build performance
   Add OpenContainers-compatible labels (#399)
+- Print receive ping moved to new functions (#400, thanks @gsnw-sebast)
 - Avoid unsigned overflow when determining the memory size to save
   response times on systems where size\_t is the same as unsigned int
   (#412 by @auerswal)
--- a/src/fping.c
+++ b/src/fping.c
@@ -263,6 +263,18 @@ char *icmp6_param_prob_str[ICMP6_PARAM_PROB_MAXCODE + 1] = {
 };
 #endif
 
+typedef struct ip_header_result {
+    int tos;
+    int ttl;
+    uint32_t otime_ms;
+    uint32_t rtime_ms;
+    uint32_t ttime_ms;
+} IP_HEADER_RESULT;
+
+IP_HEADER_RESULT default_ip_header_result() {
+    return (IP_HEADER_RESULT){-1, -1, 0x80000000U, 0x80000000U, 0x80000000U};
+}
+
 struct event;
 typedef struct host_entry {
     int i; /* index into array */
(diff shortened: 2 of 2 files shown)


src/fping.c
```diff
@@ -2058,6 +2074,177 @@ void finish()
...
+  Inputs:  struct sockaddr_storage response_addr,
+           HOST_ENTRY *h, IP_HEADER_RESULT *ip_header_res,
+           int64_t recv_time, int64_t this_reply
+
+  Description:
+
+************************************************************/
+
+void print_recv_ext(struct sockaddr_storage *response_addr, HOST_ENTRY *h, IP_HEADER_RESULT *ip_header_res, int64_t recv_time, int64_t this_reply) { 
+    if (addr_cmp((struct sockaddr *)response_addr, (struct sockaddr *)&h->saddr)) {
```
auerswal: I'd rather you did not move this part into the print function. It is the same for the normal and the JSON code path (i.e., no JSON format output) and thus duplicated.

src/fping.c
```diff
@@ -2058,6 +2074,177 @@ void finish()
...
+  Inputs:  struct sockaddr_storage response_addr,
+           HOST_ENTRY *h, IP_HEADER_RESULT *ip_header_res,
+           int64_t recv_time, int64_t this_reply
+
+  Description:
+
+************************************************************/
+
+void print_recv_ext_json(struct sockaddr_storage *response_addr, HOST_ENTRY *h, IP_HEADER_RESULT *ip_header_res, int64_t recv_time, int64_t this_reply) {
+    if (addr_cmp((struct sockaddr *)response_addr, (struct sockaddr *)&h->saddr)) {
```
auerswal: I'd rather you did not move this part into the print function. It is the same for the normal and the JSON code path (i.e., no JSON format output) and thus duplicated.

auerswal (changes requested):
I think it would be better if you could move the computations outside of the print functions. They are currently duplicated, and just moving the code into additional functions keeps them duplicated. Ideally, moving printing to additional functions should reduce code duplication.

src/fping.c
gsnw-sebast: You're right, I changed it.

src/fping.c
gsnw-sebast: You're right, I changed it.

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ Next
   (#392, thanks @gsnw-sebast and @auerswal)
 - Switch to alpine-based multi-stage Docker build to reduce image size and improve build performance
   Add OpenContainers-compatible labels (#399)
+- Print recive ping moved to new functions (#400, thanks @gsnw-sebast)
```
schweikert: nit: typo "recive"

schweikert (approved):

CHANGELOG.md
gsnw-sebast: Oh, right, I corrected that.

PR #417, opened 2025-10-23
ci: Start the GitHub action Build and Push Docker Image only when pushing to develop branch

gsnw-sebast:
When a new branch is created and a push is made to it, the GitHub action `Build and Push Docker Image` is also started.
This change means that a Docker image is only built and published when a push is made to the develop branch.

### DIFF
--- a/.github/workflows/docker-deployment.yml
+++ b/.github/workflows/docker-deployment.yml
@@ -13,7 +13,7 @@ permissions:
 env:
   REGISTRY: ghcr.io
 jobs:
-  push_to_regsitry:
+  push_to_registry:
     if: >
       github.event.workflow_run.conclusion == 'success' &&
       github.event.workflow_run.event == 'push' &&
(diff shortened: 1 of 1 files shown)


.github/workflows/docker-deployment.yml
```diff
@@ -16,7 +16,8 @@ jobs:
   push_to_regsitry:
```
auerswal: This looks like an existing typo (regsitry instead of registry) that does not seem to affect operation of this workflow.

.github/workflows/docker-deployment.yml
```diff
@@ -16,7 +16,8 @@ jobs:
...
     if: >
       github.event.workflow_run.conclusion == 'success' &&
-      github.event.workflow_run.event == 'push'
+      github.event.workflow_run.event == 'push' &&
+      github.event.workflow_run.head_branch == 'develop'
```
auerswal, line 20: This looks good.

auerswal (approved):
Looks good to me. The Docker image only needs to be updated after pushing to the `develop` branch, and this PR seems to do just that. Thanks!

.github/workflows/docker-deployment.yml
gsnw-sebast: Thanks for pointing that out, I've corrected it.

auerswal (approved):
Looks good, thanks!

PR #420, opened 2025-12-07
JSON format documentation

schweikert:
(with help from Gemini 3)

### DIFF
--- a/doc/fping-json.md
+++ b/doc/fping-json.md
@@ -1,0 +1,181 @@
+# fping JSON Output Format
+last updated for: fping 5.5
+
+> [!IMPORTANT]
+> The JSON output feature is currently in **alpha**. The format is subject to change in future versions.
+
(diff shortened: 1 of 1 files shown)


schweikert:
@gsnw-sebast Could you review? I have a few changes in mind before we release to the JSON format, which became apparent when creating this document.

1. Allow --json with --loop
2. Remove unnecessary fields in `resp`. JSON output is designed mostly for machines, and calculating averages seems unnecessary. Also, the "loss" vs "return" logic is weird for a machine-parseable format. Remove: `rttAvg`, `loss` and `return`.
3. Same for `timeout`: remove `rttAvg` and `loss`
4. `summary`: rename `outage(ms)` to `outage`.
5. Rename `globalSum` to `stats`.

I'll create a separate PR(s) for these changes.

auerswal:
Thanks for adding JSON output documentation! The documentation needs to be adjusted to the recent changes from PR #423, I think.

PR #421, opened 2025-12-07
New option --traceroute to send a traceroute with fping

gsnw-sebast:
This is my first working version for sending a traceroute with fping.

It can be tested with the following command. I have omitted IPv6 for now, and some of the text still needs to be revised.
The primary focus was on the function itself.

`./fping --traceroute 8.8.8.8`

**Output**
```
fping traceroute to 127.0.0.1 (127.0.0.1), 30 hops max, 84 byte packets
127.0.0.1: hop 1 reached DESTINATION 127.0.0.1 (0.055 ms)
```

### DIFF
--- a/src/fping.c
+++ b/src/fping.c
@@ -168,6 +168,9 @@ extern int h_errno;
 #define RESP_ERROR -3
 #define RESP_TIMEOUT -4
 
+/* Traceroute */
+#define TRACEROUTE_DONE_TTL 100
+
 /* debugging flags */
 #if defined(DEBUG) || defined(_DEBUG)
 #define DBG_TRACE 1
@@ -297,6 +300,7 @@ typedef struct host_entry {
     int64_t min_reply_i; /* shortest response time */
     int64_t total_time_i; /* sum of response times */
     int64_t *resp_times; /* individual response times */
+    int trace_ttl; /* current traceroute ttl */
 
     /* to avoid allocating two struct events each time that we send a ping, we
      * preallocate here two struct events for each ping that we might send for
@@ -423,6 +427,7 @@ int timestamp_flag = 0;
 int timestamp_format_flag = 0;
 int random_data_flag = 0;
 int cumulative_stats_flag = 0;
+int traceroute_flag = 0;
 int check_source_flag = 0;
 int icmp_request_typ = 0;
 int print_tos_flag = 0;
@@ -642,6 +647,7 @@ int main(int argc, char **argv)
         { "rdns", 'd', OPTPARSE_NONE },
         { "timestamp", 'D', OPTPARSE_NONE },
         { "timestamp-format", '0', OPTPARSE_REQUIRED },
+        { "traceroute", '0', OPTPARSE_NONE },
         { "elapsed", 'e', OPTPARSE_NONE },
         { "file", 'f', OPTPARSE_REQUIRED },
         { "generate", 'g', OPTPARSE_NONE },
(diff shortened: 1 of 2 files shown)


src/fping.c
```diff
@@ -1077,6 +1085,11 @@ int main(int argc, char **argv)
         exit(1);
     }
 
+    if (traceroute_flag && (count_flag || loop_flag)) {
+        fprintf(stderr, "%s: option --traceroute not support -c, -C, or -l\n", prog);
```
auerswal: The following might be a slightly better error message:
```
fprintf(stderr, "%s: can't combine --traceroute with -c, -C, or -l\n", prog);
```

src/fping.c
```diff
@@ -1806,11 +1830,29 @@ void main_loop()
 
             dbg_printf("%s [%d]: ping event\n", h->host, event->ping_index);
 
+            /* Traceroute */
+            if (traceroute_flag) {
+                int ttl_set = h->trace_ttl;
+                if (ttl_set > 30) ttl_set = 30;
```
auerswal: Please use a named constant instead of the number 30.
I'd think the definition should go into `src/options.h`.
I'd think it would be useful to have an option to adjust the maximum TTL / hop count value.

auerswal (commented):
Thanks for working on this, @gsnw-sebast!
At least on Linux, receiving ICMP error messages requires the use of raw sockets. Please add checks that `fping` opened raw sockets when using `--traceroute`.
I have not yet looked at all code in detail, this was a first glance.

src/fping.c
gsnw-sebast: Thanks, I've already changed that locally. I'm still testing and considering which options are still unnecessary or whether I should disable all of them for now.

src/fping.c
gsnw-sebast: I wanted to add that. For testing purposes, I set it to the maximum value of 30.
In general, we have the option `-H, --ttl set the IP TTL value (Time To Live hops)`.

But I still have to see how I can implement this exactly, because only values between 1 and 30 should be allowed for traceroute.

gsnw-sebast:
Checklist with options that allow and tested
| Option | Test | Supported |
| ---------- | ------ | --------------- |
| --ipv4 (-4) |  |  |
| --ipv6 (-6) |  |  |
| --alive (-a) | Yes | No (No effect) |
| --addr (-A) | Yes | Yes |
| --size (-b) | Yes | Yes (I couldn't really check the results and effects.) |
| --backoff (-B) | Yes | No (No effect) |
| --count (-c) | Yes  | No  |
| --vcount (-C) | Yes  | No |
| --rdns (-d) | Yes | Yes |
| --timestamp (-D) |  |  |
| --elapsed (-e) | Yes | No (No effect) |
| --file (-f) |  |  |
| --generate (-g) |  |  |
| --ttl (-H) | Yes  | Yes (1-30) |
| --interval (-i) | Yes | Yes |
| --iface (-I) | No | Yes (Untested) |
| --json (-j) |  |  |
| --icmp-timestamp |  |  |
| --fwmark (-k) | No | No (Untested) |
| --loop (-l) | Yes  | No |
| --all (-m) | Yes | No (No effect) |
| --dontfrag (-M) | Yes | Yes |
| --name (-n) | Yes | Yes |
| --netdata (-N) | Yes  | No |
| --outage (-o) |  |  |
| --tos (-O) | Yes  | Yes |
| --period (-p) | Yes | Yes (same effect as -i) |
| --quiet (-q) | Yes | No |
| --squiet (-Q) | Yes | No |
| --retry (-r) | Yes | No (No effect) |
| --random (-R) | Yes | Yes |
| --stats (-s) | Yes | No (Works, but the output is misleading) |
| --src (-S) | No | Yes (Untested) |
| --timeout (-t) | Yes | No (No effect) |
| (-T) |  |  |
| --unreach (-u) | Yes | No (No effect)  |
| --retry (-r) | Yes | No (No effect)  |
| --fast-reachable (-X) |  |  |
| --check-source |  |  |
| --print-tos | Yes | No |
| --print-ttl | Yes | No |
| --seqmap-timeout |  |  |
| (-z) |  |  |

PR #422, opened 2025-12-07
Allow --json also for loop-mode

schweikert:


gsnw-sebast (approved):
That's fine with me, I just tested it and it works just like without JSON output.

But now I'm wondering why I left it out back then?!?

PR #423, opened 2025-12-07
json: simplify recv and timeout output (and code)

schweikert:
The rttAvg, loss, and return fields are not really useful IMO, at least not in the JSON output, which is mainly targeting machine-processing of the output (and calculating averages / percentages as you see fit, is easy enough).


gsnw-sebast (approved):
Thanks for the correction. It looks much more elegant, and I don't mind the fields being omitted. If they are needed, the user should create a ticket with an explanation and we will reinstall them.

PR #439, opened 2025-12-25
Improve robustness of downloading of autotools tarballs

schweikert:


gsnw-sebast:
This is due to unstable servers with HTTPS.
Try HTTP instead of HTTPS.

I had the same problem with another project on October 3, 2025, with https://ftp.gnu.org/gnu/bison/
After switching to http://ftp.gnu.org/gnu/bison/, I never had any problems again.

schweikert:
> This is due to unstable servers with HTTPS. Try HTTP instead of HTTPS.
> 
> I had the same problem with another project on October 3, 2025, with https://ftp.gnu.org/gnu/bison/ After switching to http://ftp.gnu.org/gnu/bison/, I never had any problems again.

Ah thanks for the tip! I changed it now to HTTP.

gsnw-sebast:
Change back to http://ftp.gnu.org/gnu/ that works

PR #440, opened 2025-12-25
GitHub actions: don't run tests twice for github PRs

schweikert:


gsnw-sebast (changes requested):
With this change, the test would only run with pull to develop or with a pull request.
I would suggest allowing manual execution via “workflow_dispatch” as well, so that fork repositories can also execute the job independently of pull requests and push to develop.

schweikert:
> With this change, the test would only run with pull to develop or with a pull request. I would suggest allowing manual execution via “workflow_dispatch” as well, so that fork repositories can also execute the job independently of pull requests and push to develop.

Done, thanks

PR #444, opened 2025-12-28
Fix usage pod2man under FreeBSD and OpenBSD

gsnw-sebast:
When building fping on FreeBSD and OpenBSD, the process hangs at doc/Makefile.in. With this change, the whole thing works even without gmake or the OpenBSD patch.

OpenBSD patch: https://cvsweb.openbsd.org/ports/net/fping/patches/patch-doc_Makefile_in

### DIFF
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -3,4 +3,4 @@ man_MANS = fping.8
 EXTRA_DIST = fping.8 fping.pod README.1992 CHANGELOG.pre-v4
 
 fping.8: fping.pod
-	pod2man  -c "" -s 8 -r "fping" $< >$@
+	pod2man  -c "" -s 8 -r "fping" fping.pod > fping.8


configure.ac
```diff
@@ -10,19 +10,24 @@ m4_ifdef([AC_AUTOCONF_VERSION],[AC_USE_SYSTEM_EXTENSIONS], [AC_GNU_SOURCE])
...
     only_clock_realtime=yes
     ;;
   *freebsd*)
     only_clock_realtime=yes
+    os_is_bsd=yes
     ;;
   *openbsd*)
     only_clock_realtime=yes
+    os_is_bsd=yes
     ;;
```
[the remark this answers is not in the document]
gsnw-sebast: This is kept separate to make later adjustments easier.

doc/Makefile.am
```diff
@@ -2,5 +2,10 @@ man_MANS = fping.8
 
 EXTRA_DIST = fping.8 fping.pod README.1992 CHANGELOG.pre-v4
 
+if OS_IS_BSD
```
schweikert: This doesn't seem worth the hassle. Let's just use "fping.pod" instead of $> and $<?

gsnw-sebast: I undid everything and entered it statically.

schweikert (approved):

PR #446, opened 2026-01-01
ci: Removing travis-ci

gsnw-sebast:
Removing travis-ci because it is no longer actively used and travis-ci does not offer a free option for opensource projects


schweikert (approved):

gsnw-sebast:
I'm not sure, but do we need ci/deploy-coverity.sh or did that only running with travis-ci?

schweikert:
> I'm not sure, but do we need ci/deploy-coverity.sh or did that only running with travis-ci?

I would keep that script. It probably can be ported to run using Github Actions.

schweikert:
I fixed the coverity stuff (it now runs with GitHub Actions)

PR #461, opened 2026-02-14
Option to show used IP source address in output

### LINKED ISSUE #261: Option to show used IP source address in output
Hi,

I'm asking for an option to show the source address used when running fping.

While it's possible to define a source address/interface, it would be interesting to know which source address is used when no source is specified. As the pings always go between two IP addresses, I would like to know which source is used.

Could such an option be added? That would be very nice!

Thanks,
Simon

gsnw-sebast:
New function for outputting the source IP address.
IPv6 is currently not supported. Major changes are necessary here.
Under Ubuntu 22.04, the test `./fping --print-srcaddr 127.0.0.1` always returns `unknown`.

Issues: #261


ci/test-16-json-output.pl
```diff
@@ -98,6 +98,13 @@
 $cmd->stderr_is_eq("");
 }
 
+# fping -J -c 1 --print-srcaddr 127.0.0.1
+my $cmd = Test::Command->new(cmd => "fping -J -c 1 --print-srcaddr 127.0.0.1");
+$cmd->exit_is_num(0);
+$cmd->stdout_like(qr/^\{"resp":\s\{"host":\s"127\.0\.0\.1",\s"seq":\s0,\s"size":\s\d+,\s"rtt":\s\d+\.\d+,\s"src":\s"\d+\.\d+\.\d+\.\d+"\}\}
+\{"summary":\s\{"host":\s"127\.0\.0\.1",\s"xmt":\s\d+,\s"rcv":\s\d+,\s"loss":\s\d+,\s"rttMin":\s\d+\.\d+,\s"rttAvg":\s\d+\.\d+,\s"rttMax":\s\d+\.\d+\}\}\n?$/);
```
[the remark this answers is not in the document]
gsnw-sebast: Under Ubuntu 22.04, the test ./fping --print-srcaddr 127.0.0.1 always returns unknown.
I tested this with ubuntu-latest on Azure DevOps, and the behavior does not occur there.

gsnw-sebast:
I've already finished the version with IPv6, but I'll submit it in a separate pull request.

schweikert:
Could you update fping-json.md too?

gsnw-sebast:
I just added that. Thanks for the tip

ci/test-11-unpriv.pl
```diff
@@ -40,14 +40,20 @@ sub get_ping_gid_range {
...
     {
         my $cmd = Test::Command->new(cmd => "$fping_copy 127.0.0.1");
         $cmd->exit_is_num(0);
         $cmd->stdout_is_eq("127.0.0.1 is alive\n");
         $cmd->stderr_is_eq("");
     }
+    {
+        my $cmd = Test::Command->new(cmd => "$fping_copy --print-reply-dst 127.0.0.1");
+        $cmd->exit_is_num(0);
+        $cmd->stdout_like(qr{127\.0\.0\.1 is alive \(Reply-DST (?:\d+\.\d+\.\d+\.\d+|unknown)\)\n});
```
auerswal: This test must always result in `unknown`, because the code uses the IPv4 header which is only available with IPv4 raw sockets. IPv4 raw sockets require privileges. Please tighten the test to require `unknown`.
If we should add support for unprivileged operation later (this works similar as the IPv6 case), then we should change the test to return an IPv4 address instead of `unknown`, probably to return `127.0.0.1` since we know which IPv4 address is used to ping the loopback address `127.0.0.1`.

ci/test-16-json-output.pl
```diff
@@ -98,6 +98,13 @@
 $cmd->stderr_is_eq("");
 }
 
+# fping -J -c 1 --print-srcaddr 127.0.0.1
+my $cmd = Test::Command->new(cmd => "fping -J -c 1 --print-reply-dst 127.0.0.1");
+$cmd->exit_is_num(0);
+$cmd->stdout_like(qr/^\{"resp":\s\{"host":\s"127\.0\.0\.1",\s"seq":\s0,\s"size":\s\d+,\s"rtt":\s\d+\.\d+,\s"replyDst":\s"\d+\.\d+\.\d+\.\d+"\}\}
```
auerswal: Can this be tightened to require `127.0.0.1` as the reply's destination address?

auerswal (approved):
In general, this looks OK to me. I'd like you to look into tightening the tests a bit, but I don't insist.

ci/test-11-unpriv.pl
gsnw-sebast: I've made the correction, thanks for pointing that out.

ci/test-16-json-output.pl
gsnw-sebast: I've made the correction, thanks for pointing that out.

PR #466, opened 2026-03-13
Add IBM i PASE detection

### LINKED ISSUE #465: compile in PASE IBM i from sources
Hello, I've downloaded 5.5 and compiled on a PASE IBM i environment (it's like an AIX POSIX environment , power processor , inside a IBM i system).

Just to inform you, in order to compile it via gcc, I had to define these variables during configure

./configure --prefix=/QOpenSys/usr CC=/QOpenSys/pkgs/bin/gcc CPPFLAGS="-I/QOpenSys/pkgs/include -D_SIGSET_T -DHAVE_STDLIB_H -DIP_RECVTOS=0" CFLAGS="-g -O2" 

_SIGSET_T required due to avoid a redefinition (guard) in a double include h on the system.
HAVE_STDLIB_H forced because it wasn't included by the auto configuration.
IP_RECVTOS forced to let it compile (this socket options is missing in this environment like some BSDs).

I just note here in case someone will find this useful on PASE.

BTW such gcc I see define the variable "\_\__PASE\_\_" when run in PASE environment... maybe it is possible to include in the stock sources a conditional def that auto include such vars if \_\_PASE\_\_ is defined? just to streamline the compilation process for this useful tool.
thanks

gsnw-sebast:
Enabling detection of IBM i systems during source compilation
This was requested in issue #465

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ Next
   function declarations to a separate file (#454, thanks @gsnw-sebast)
 - Avoid crash with option `-0` (#457, thanks @auerswal)
 - Move option parsing to a separate file fping.c and fping.h (#458, thanks @gsnw-sebast)
+- Add IBM i PASE detection (#466, thanks @acerbis and @gsnw-sebast)
 - ci: Update autoconf to version 2.73 (#467, thanks @gsnw-sebast)
 
 fping 5.5 (2025-12-31)
--- a/configure.ac
+++ b/configure.ac
@@ -23,6 +23,31 @@ case "${target}" in
     ;;
 esac
 
+dnl Detect IBM i PASE environment
+AC_MSG_CHECKING([for IBM i PASE environment])
+AC_COMPILE_IFELSE(
+  [AC_LANG_PROGRAM([[]], [[
+#if !defined(__PASE__)
+#error not PASE
+#endif
+  ]])],
+  [
+    AC_MSG_RESULT([yes])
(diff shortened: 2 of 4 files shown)


gsnw-sebast:
@acerbis
I've implemented the improvements suggested by gemini-code-assist.
Can you test again to see if this works for IBM i?

acerbis:
Does **NOT** work now (was working in the first attempt, just move the var defs before, prepending and not post).

-DHAVE_STDLIB_H is necessary because configure doesn't detect it apparently autonomously so compilation will fail.
-D_SIGSET_T is necessary to avoid double definitions (guard).
-DIP_RECVTOS=0 is necessary because some platforms doesn't have a setsockoptions possible with TOS

yes -DIP_RECVTOS=0 is clearly hackish but doesn't have big impact.

A more elegant way would be to have in fping a conditional to skip TOS releated code stuff for the platforms lacking TOS handling and then use such variable if TOS is present or not (i.e. like you did with __PASE__ try in .ac to compile a source with IP_RECVTOS if the symbol is missing one would disable TOS support it).

my 2c

gsnw-sebast:
I've put the old code back in, but the workaround using `-DIP_RECVTOS=0` has been resolved here in the code.
Please test again.

The test scripts in the `ci/` folder fail on IBM i when --print-tos is included in the command.

acerbis:
Hello!
I confirm working. Thanks.

schweikert (approved):

PR #470, opened 2026-05-17
Fix getprotobyname missing on Android preventing use

Paliak:
`getprotobyname` is [stubbed out on Android](https://android.googlesource.com/platform/bionic/+/089111a80a6a167de827bf32ae2d2a935ff5874f%5E2..089111a80a6a167de827bf32ae2d2a935ff5874f/) causing crash_and_burn to be called. This commit removes the ICMP support check when compiling for Android and adds a configuration option to compile without `getprotobyname` check.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -10,7 +10,7 @@ m4_ifdef([AC_AUTOCONF_VERSION],[AC_USE_SYSTEM_EXTENSIONS], [AC_GNU_SOURCE])
 # Detect Operatingsystem
 AC_CANONICAL_TARGET
 only_clock_realtime=no
-use_getprotobyname=yes
+disable_getprotobyname=no
 
 case "${target}" in
   *darwin*)
@@ -23,7 +23,7 @@ case "${target}" in
     only_clock_realtime=yes
     ;;
   *android*)
-    use_getprotobyname=no
+    disable_getprotobyname=yes
     ;;
 esac
 
@@ -108,7 +108,20 @@ AC_ARG_ENABLE([debug],
 AS_IF([test "x$enable_debug" = "xyes"], [
   AC_DEFINE([DEBUG], [1], [Define if debugging is enabled])])
 
-AS_IF([test "x$use_getprotobyname" = "xyes"], [AC_DEFINE([USE_GETPROTOBYNAME], [1], [Enable use of getprotobyname])])
+AS_IF([test "x$only_clock_realtime" = "xyes"], [AC_DEFINE(ONLY_CLOCK_REALTIME, [1], [ONLY_CLOCK_REALTIME is defined])])
+
+dnl check for getprotobyname support
+
+AC_CHECK_FUNCS([getprotobyname])
+
+AS_IF([test "x$disable_getprotobyname" = "xyes"], [
+  AC_MSG_NOTICE([Android detected: Disabling getprotobyname])
+  ac_cv_func_getprotobyname=no
+])
+
+if test $ac_cv_func_getprotobyname = yes; then
+  AC_DEFINE([USE_GETPROTOBYNAME], [1], [Define if getprotobyname is available and should be used.])
+fi
 
 AM_INIT_AUTOMAKE([-Wall -Werror foreign])
 AM_MAINTAINER_MODE
--- a/configure.ac
+++ b/configure.ac
@@ -104,6 +104,14 @@ AC_ARG_ENABLE([debug],
 AS_IF([test "x$enable_debug" = "xyes"], [
   AC_DEFINE([DEBUG], [1], [Define if debugging is enabled])])
 
+
+AC_ARG_ENABLE([getprotobyname],
+  AS_HELP_STRING([--disable-getprotobyname], [Disable use of getprotobyname]),
+  [], [enable_getprotobyname=yes])
+AS_IF([test "x$enable_getprotobyname" = "xyes"], [
+  AC_DEFINE([USE_GETPROTOBYNAME], [1], [Enable use of getprotobyname])
+])
+
 AM_INIT_AUTOMAKE([-Wall -Werror foreign])
 AM_MAINTAINER_MODE
 
--- a/configure.ac
+++ b/configure.ac
@@ -10,7 +10,6 @@ m4_ifdef([AC_AUTOCONF_VERSION],[AC_USE_SYSTEM_EXTENSIONS], [AC_GNU_SOURCE])
 # Detect Operatingsystem
 AC_CANONICAL_TARGET
 only_clock_realtime=no
-use_getprotobyname=yes
 
 case "${target}" in
   *darwin*)
@@ -21,9 +20,6 @@ case "${target}" in
     ;;
   *openbsd*)
     only_clock_realtime=yes
-    ;;
-  *android*)
-    use_getprotobyname=no
     ;;
 esac
 
@@ -107,8 +103,6 @@ AC_ARG_ENABLE([debug],
   AS_HELP_STRING([--enable-debug], [enable debugging @<:@default=no@:>@]), [enable_debug=$enableval], [enable_debug=no])
 AS_IF([test "x$enable_debug" = "xyes"], [
   AC_DEFINE([DEBUG], [1], [Define if debugging is enabled])])
-
-AS_IF([test "x$use_getprotobyname" = "xyes"], [AC_DEFINE([USE_GETPROTOBYNAME], [1], [Enable use of getprotobyname])])
 
 AM_INIT_AUTOMAKE([-Wall -Werror foreign])
 AM_MAINTAINER_MODE
--- a/configure.ac
+++ b/configure.ac
@@ -10,6 +10,7 @@ m4_ifdef([AC_AUTOCONF_VERSION],[AC_USE_SYSTEM_EXTENSIONS], [AC_GNU_SOURCE])
 # Detect Operatingsystem
 AC_CANONICAL_TARGET
 only_clock_realtime=no
+use_getprotobyname=yes
 
 case "${target}" in
   *darwin*)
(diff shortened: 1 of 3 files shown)


src/socket4.c
```diff
@@ -59,20 +59,23 @@ static int outgoing_src_addr_set_ipv4 = 0;
...
+    int p_proto = IPPROTO_ICMP;
 
-    /* confirm that ICMP is available on this machine */
-    if ((proto = getprotobyname("icmp")) == NULL)
-        crash_and_burn("icmp: unknown protocol");
+#if defined(USE_GETPROTOBYNAME) && !(defined(ANDROID) || defined(__ANDROID__))
+	/* confirm that ICMP is available on this machine */
+	if (getprotobyname("icmp") == NULL) {
+		crash_and_burn("icmp: unknown protocol");
+	}
```
gsnw-sebast: On Linux or Unix, you should continue to use `getprotobyname(‘icmp’)`. Your change hard codes the value to `IPPROTO_ICMP`.

Perhaps the following would be better?

```
struct protoent* proto = getprotobyname("icmp");
if (proto == NULL) {
    crash_and_burn("icmp: unknown protocol");
}
p_proto = proto->p_proto;
```

configure.ac
```diff
@@ -104,6 +104,14 @@ AC_ARG_ENABLE([debug],
 AS_IF([test "x$enable_debug" = "xyes"], [
   AC_DEFINE([DEBUG], [1], [Define if debugging is enabled])])
 
+
+AC_ARG_ENABLE([getprotobyname],
+  AS_HELP_STRING([--disable-getprotobyname], [Disable use of getprotobyname]),
```
gsnw-sebast: Is the `--disable-getprotobyname` flag really necessary? If not, then an attempt should be made to perform an automatic check.
For example, using `AC_CHECK_FUNCS()`

src/socket4.c
Paliak: I thought about doing that and wasn't sure how much value getting p_proto through getprotobyname offers. To my understanding `IPPROTO_ICMP` should always (?) be the same as the return value of `getprotobyname("icmp")`.

Don't mind either way.

configure.ac
Paliak: This was the most portable way i could think of implementing this. `AC_CHECK_FUNCS` will return true on Android as `getprotobyname` does exist. [It just always returns NULL](https://android.googlesource.com/platform/bionic/+/089111a80a6a167de827bf32ae2d2a935ff5874f%5E2..089111a80a6a167de827bf32ae2d2a935ff5874f/). The only automatic way to check this that i know of would be to use `AC_RUN_IFELSE` but i decided against that as i feel it would be quite messy (and maybe cause issues with cross compilation). I guess the flag could just be removed altogether as `#if !(defined(ANDROID) || defined(__ANDROID__))` would suffice for what this PR is trying to accomplish.

Just thought that it could be a nice to have.

gsnw-sebast: In that case, it would be better to move the Android detection to `configure.ac`. We already have the `case "${target}" in` function, and if other systems use the same approach, it can be extended quite easily.
This keeps the exception in the code itself more flexible.

src/socket4.c
gsnw-sebast: I'm not aware of any such system at the moment, but perhaps there are some exotic systems where this isn't the case, and it would represent a break with the POSIX/network abstraction layer

configure.ac
Paliak: Implemented in [be8cc59](https://github.com/schweikert/fping/pull/470/commits/be8cc590df6dceca467a0770e7054fdb72dbdda1) let me know what you think.

src/socket4.c
Paliak: Implemented your suggestion in [0d09161](https://github.com/schweikert/fping/pull/470/commits/0d09161499cf8b1de1e26a1c8c328f53b67eefd9).

configure.ac
```diff
@@ -10,6 +10,7 @@ m4_ifdef([AC_AUTOCONF_VERSION],[AC_USE_SYSTEM_EXTENSIONS], [AC_GNU_SOURCE])
 # Detect Operatingsystem
 AC_CANONICAL_TARGET
 only_clock_realtime=no
+use_getprotobyname=yes
```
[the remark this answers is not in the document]
Paliak: Not applicable as per https://github.com/schweikert/fping/pull/470#discussion_r3326558167

configure.ac
```diff
@@ -21,6 +22,9 @@ case "${target}" in
   *openbsd*)
     only_clock_realtime=yes
     ;;
+  *android*)
+    use_getprotobyname=no
+    ;;
```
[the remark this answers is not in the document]
Paliak: Not applicable as per https://github.com/schweikert/fping/pull/470#discussion_r3326558167

configure.ac
```diff
@@ -104,6 +108,8 @@ AC_ARG_ENABLE([debug],
 AS_IF([test "x$enable_debug" = "xyes"], [
   AC_DEFINE([DEBUG], [1], [Define if debugging is enabled])])
 
+AS_IF([test "x$use_getprotobyname" = "xyes"], [AC_DEFINE([USE_GETPROTOBYNAME], [1], [Enable use of getprotobyname])])
```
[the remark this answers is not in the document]
Paliak: Not applicable as per https://github.com/schweikert/fping/pull/470#discussion_r3326558167

src/socket4.c
```diff
@@ -59,20 +59,27 @@ static int outgoing_src_addr_set_ipv4 = 0;
...
 {
-    struct protoent* proto;
-    int s;
+    int s = -1;
+    int p_proto = IPPROTO_ICMP;
 
-    /* confirm that ICMP is available on this machine */
-    if ((proto = getprotobyname("icmp")) == NULL)
-        crash_and_burn("icmp: unknown protocol");
+#ifdef USE_GETPROTOBYNAME
```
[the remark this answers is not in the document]
Paliak: Not applicable as per https://github.com/schweikert/fping/pull/470#discussion_r3326558167

src/socket6.c
```diff
@@ -57,20 +57,27 @@ static int outgoing_src_addr_set_ipv6 = 0;
...
 {
-    struct protoent* proto;
-    int s;
+    int s = -1;
+    int p_proto = IPPROTO_ICMPV6;
 
-    /* confirm that ICMP6 is available on this machine */
-    if ((proto = getprotobyname("ipv6-icmp")) == NULL)
-        crash_and_burn("ipv6-icmp: unknown protocol");
+#ifdef USE_GETPROTOBYNAME
```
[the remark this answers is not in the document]
Paliak: Not applicable as per https://github.com/schweikert/fping/pull/470#discussion_r3326558167

configure.ac
gsnw-sebast: I would first check whether `getprotobyname` is present, and then check whether it needs to be disabled again.

Here's my example:
```
# Detect Operatingsystem
AC_CANONICAL_TARGET
only_clock_realtime=no
disable_getprotobyname=no

case "${target}" in
  *darwin*)
    only_clock_realtime=yes
    ;;
  *freebsd*)
    only_clock_realtime=yes
    ;;
  *openbsd*)
    only_clock_realtime=yes
    ;;
  *android*)
    disable_getprotobyname=yes
    ;;
esac

dnl Checks for libraries.

AC_CHECK_FUNCS([getprotobyname])

AS_IF([test "x$disable_getprotobyname" = "xyes"], [
  AC_MSG_NOTICE([Android detected: Disabling getprotobyname])
  ac_cv_func_getprotobyname=no
])

if test $ac_cv_func_getprotobyname = yes; then
  AC_DEFINE([USE_GETPROTOBYNAME], [1], [Define if getprotobyname is available and should be used.])
fi
```

Paliak: Sorry for the late reply. I've applied your suggestion. Good idea on checking for getprotobyname before disabling it for android specifically. Should help with support for other platforms where the function may be missing, not just android that i'm focused on.

src/socket4.c
```diff
@@ -59,20 +59,27 @@ static int outgoing_src_addr_set_ipv4 = 0;
 
 int open_ping_socket_ipv4(int *socktype)
 {
-    struct protoent* proto;
-    int s;
+    int s = -1;
+    int p_proto = IPPROTO_ICMP;
```
schweikert: Can't we just use this on all platforms? It seems to me that maybe we don't need getprotobyname at all.

schweikert (changes requested):

src/socket4.c
Paliak: [That was my thinking too](https://github.com/schweikert/fping/pull/470#discussion_r3326507563). @gsnw-sebast suggested there may be situations where `IPPROTO_ICMP` differs from the return value of `getprotobyname`, so I added it back per that suggestion.

If we really want to check whether ICMP is supported on a given platform maybe we could just check for the define?

gsnw-sebast: I looked into it, and you're actually right. The value could be set statically without the check via `getprotobyname()`. The numbers are assigned by the IANA [1], and even if fping were to support TCP or UDP in the future, these numbers would remain unchanged with `TCP = 6` and `UDP = 17`.

It would even keep fping running if an incorrect protocol file were stored on the system

[1] https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml

Paliak: Removed `getprotobyname` altogether in [4cd0304](https://github.com/schweikert/fping/pull/470/commits/4cd0304af30d5e827dfa9dfeb4e4b80569598ac9).

gsnw-sebast (approved):
As far as I'm concerned, we can go ahead with that. We just need to update the changelog and prepare for a rebase merge.

sollybr:
Any ETA on merge?

schweikert (approved):

